PR #41, opened 2022-04-16
feat: bind workspaces to specific monitors

lyze237:
This is the only PR im not 100% happy with as the DisplayNames are quite generic (e.g. `\\.\DISPLAY1`).

Ideally in my opinion, it'd be cool to have the actual monitor names as usable strings on the config. I've googled around a bit and apparently it's quite hard to figure that out, so I haven't tried to implement that.

### DIFF
--- a/GlazeWM.Domain/Monitors/CommandHandlers/AddMonitorHandler.cs
+++ b/GlazeWM.Domain/Monitors/CommandHandlers/AddMonitorHandler.cs
@@ -49,9 +49,9 @@ namespace GlazeWM.Domain.Monitors.CommandHandlers
 
     private void ActivateWorkspaceOnMonitor(Monitor monitor)
     {
-      // Get name of first workspace that is not active.
-      var inactiveWorkspaceName =
-        _workspaceService.GetInactiveWorkspaceNames().ElementAtOrDefault(0);
+      // Get name of first workspace that is not active for that specified monitor or any.
+      var inactiveWorkspaceName = _workspaceService.GetInactiveWorkspaceNameForMonitor(monitor) ??
+                                  _workspaceService.GetInactiveWorkspaceNames().ElementAtOrDefault(0);
 
       if (inactiveWorkspaceName == null)
         throw new FatalUserException("At least 1 workspace is required per monitor.");
--- a/GlazeWM.Domain/Monitors/CommandHandlers/AddMonitorHandler.cs
+++ b/GlazeWM.Domain/Monitors/CommandHandlers/AddMonitorHandler.cs
@@ -51,7 +51,7 @@ namespace GlazeWM.Domain.Monitors.CommandHandlers
     {
       // Get name of first workspace that is not active for that specified monitor or any.
       var inactiveWorkspaceName = _workspaceService.GetInactiveWorkspaceNameForMonitor(monitor) ??
-                                  _workspaceService.GetInactiveWorkspaceNames().ElementAtOrDefault(0);
+                                  _workspaceService.GetInactiveWorkspaceNamesNotDedicatedToAMonitor().ElementAtOrDefault(0);
 
       if (inactiveWorkspaceName == null)
         throw new FatalUserException("At least 1 workspace is required per monitor.");
--- a/GlazeWM.Domain/Monitors/CommandHandlers/AddMonitorHandler.cs
+++ b/GlazeWM.Domain/Monitors/CommandHandlers/AddMonitorHandler.cs
@@ -51,7 +51,8 @@ namespace GlazeWM.Domain.Monitors.CommandHandlers
     {
       // Get name of first workspace that is not active for that specified monitor or any.
       var inactiveWorkspaceName = _workspaceService.GetInactiveWorkspaceNameForMonitor(monitor) ??
-                                  _workspaceService.GetInactiveWorkspaceNamesNotDedicatedToAMonitor().ElementAtOrDefault(0);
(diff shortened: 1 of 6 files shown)


lars-berger:
Haven't done a lot of testing when it comes to this stuff to be honest. Does the `DeviceName` stay the same after restarts? If they don't, it'd be pretty annoying to change the config every time. New monitor connections and disconnections might also complicate things.

Also curious whether `\\.\DISPLAY{X}` always maps to the numbers in the display settings screens (eg. `\\.\DISPLAY1` -> `1`). If that's the case, and if the `DeviceName` is actually consistent, then it could be viable to simply set something like `bind_to_monitor : 1`. It'd also be way easier for users to figure out the IDs of their monitors.

image

Like you mentioned, using the name of the monitor could also be an option. I guess you mean the model name (eg. `DELL U2715H`)? Not sure what would happen if you have 2 of the same monitor though. There's also the case where users are on a laptop and there's no external display devices.

lyze237:
Yeah I was talking about the name windows reports under "display information":
image

Good thinking though about having two monitors with the same name, so I guess that's not practicable.

From what I know the number should persist between restarts as it's bound to the port of the GPU. So the only problem with that is when you swap out the monitors (e.g. on a Laptop with multiple screens when using a different port combination) that there might be different numbers but on a desktop it should be hopefully fine.

KiLLeRRaT:
It looks like the `DeviceName` will be reasonably consistent.  Between reboots at least.

Check out:
https://stackoverflow.com/questions/10669024/identifying-displays where the guy was quite happy with using `DeviceName` for his purpose.

and, https://social.msdn.microsoft.com/Forums/sqlserver/en-US/58e89043-954e-4f30-8e3e-1357fcb0550c/windows-form-application-identify-the-displays?forum=winforms

I think that if it changes with graphics driver changes, or when you plug monitors in and out, that it's fine to have to potentially have to update the config.

Initially when you start the program, if a screen is missing, we could stay that just attach those workspaces to your primary instead.  That would solve the problem if you say use a laptop with, and without an external monitor.  When you're on the road and don't have an external monitor, all those workspaces will now appear on your primary.  When you plug the monitor back in, they should go back to your external monitor when it's available.  We may need to detect a screen change event and update the workspaces when that happens.

Thoughts?

I'm quite keen on this feature, so if there is something I could help with, let me know and I can have a go at it.

Cheers,

KiLLeRRaT:
Hi @lyze237 ,

I'm trying to get this going in one of my own feature branches (because I've added a few PRs that hasn't yet been completed).

I'm struggling to get this going, or maybe I'm missing the point of this.

I have the following config:

```
workspaces:
  - name: 1
    display_name: "1:Chat"
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY1'
  - name: 2
    display_name: "2:Code"
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY2'
  - name: 3
    display_name: "3:VS"
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY3'
  - name: 4
    display_name: "4:Test"
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY1'
  - name: 5
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY2'
  - name: 6
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY3'
  - name: 7
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY1'
  - name: 8
    display_name: "8:Email"
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY2'
  - name: 9
    display_name: "9:Search"
    keep_alive: true
    bind_to_monitor: '\\.\DISPLAY3'
```

when I start the app, it shows 1 space on each of my 3 displays.  But when I switch to another space, that space opens on the monitor I'm on currently, and not on the bound monitor as per config.  Is that what this PR is meant to do?

Am I missing something somewhere?

I've also tried it without my `keep_alive` settings, and same behaviour.

You can try my branch over here if you have a minute: https://github.com/KiLLeRRaT/GlazeWM/tree/feature/20220714_ag_bindworkspacetomonitor

I'm really looking forward to get it so that workspace x always stays on monitor y, that would be awesome, so that I don't have to do all this manual setup when I start the app.

Cheers,

stewart-james:
One pain I have with GlazeWM is the inability to bind workspaces to specific monitors, in my case I use a single ultrawide monitor at home whereas 2x 1080p monitors at work. It'd be nice if we could control which monitor the workspaces display on.

One area `bind_to_monitor` (& also `\\.\Display#` ) wouldn't support is when you are moving a laptop between different monitor configurations. Perhaps the config option could be `preferred_monitors` and you provide a priority list of monitor identifiers (not sure what that identifier should be). If none of the preferred monitors could be found it falls back to the main monitor. 

With this PR I could also just have 2 copies of my config.yaml, one for my office setup and one for my home setup, using a script to activate the correct one (providing a config.yaml through the command line would be very nice, then it's just a windows shortcut required!)

lyze237:
Hey there!
Sorry for disappearing for a couple months, was quite busy doing other things.

I've updated the branch with the latest version of the repo and also used it for a while and it's as stable as it can get!

For the config file I just needed to do:
```yml
workspaces:
  - name: 1
    bind_to_monitor: \\.\DISPLAY1
  - name: 2
    bind_to_monitor: \\.\DISPLAY2
  - name: 3
    bind_to_monitor: \\.\DISPLAY3
```
and it worked for me that way.

lyze237:
Ah I actually did just that, haha!

Additionally to that I made sure that the monitor itself doesn't steal a workspace assigned to a different monitor.

lyze237:
Could you tell me where you left those comments? I can't find em.

PR #42, opened 2022-04-16
chore: housekeeping! update to .NET 6 and fix common rules violations

neon-sunset:
This project has briefly appeared on a front page of a certain orange website. Therefore, I thought it could use some love in the form of simplest and, arguably, low-effort fixes such as
- Updating it to the latest LTS target, namely, .NET 6
- Going over project files and applying fixes to the violations of MSFT and Roslynator analyzer rules
- Further minor tweaks when automatic fixes are insufficient (e.g. `BarViewModel .CreateComponentViewModels`)

All the changes are made to improve performance / fix rule violations only and intended to not introduce any changes to the functional behavior.
Updated LINQ calls (e.g. `Find` instead of `FirstOrDefault`) yield same results at lower cost because certain types (array, List<T>, etc.) have specialized versions implemented for them.


lars-berger:
This is seriously impressive, thank you!

Would love to be able to run the rule fixer in the future. Is it the [Roslynator CLI](https://github.com/JosefPihrt/Roslynator/blob/master/docs/cli/README.md) that I should be looking into?

I hate to ask for a follow-up change, but I happened to type `_lockObj` ([L15 of `Bus.cs`](https://github.com/neon-sunset/GlazeWM/blob/feat/update-to-net6/GlazeWM.Infrastructure/Bussing/Bus.cs#L15)) to be static when it shouldn't be. After the rule fix, this makes `Bus.Invoke` static and it's called everywhere throughout the app. It'd be nice to keep the change in the same MR I reckon.

neon-sunset:
Sure, I will go over `Bus` and its usage throughout the app to correctly fix it :)

As for the question about `Roslynator` - it is offered in multiple forms such as CLI you mentioned, but there are also Visual Studio and VS Code extensions. I'm using the latter at work.

In general, I think this codebase has a few issues regarding inconsistent style and some things could've been written in a cleaner way but I generally try to avoid talking about code style in OSS projects because doing tends to harm the interest of authors in the project.

I would like to submit a few follow-up PRs to address those and clean up some parts of the implementation but only if you are okay with it :)

lars-berger:
Of course, you're more than welcome to fire away some follow-up PRs. C# isn't my main language, so there's bound to be some weird code smells here and there.

Cool to hear Roslynator is also a VSCode extension, will give it a shot. Seems like a lot of C# tooling is exclusively for ReSharper or VS, so I'm glad it's an exception.

Also, not sure if this is an issue on my end, but I had to change `"program"` in `launch.json` to be able to run your fork.

```json
      "program": "${workspaceFolder}/GlazeWM.Bootstrapper/bin/Debug/net6.0-windows/GlazeWM.dll",
```

neon-sunset:
Yeah, I missed this one too, sorry. Will fix.

Since .NET Core was introduced, the ecosystem has come a long way. Visual Studio and Rider while powerful are by no means a requirement. Support for C# in VS Code and other text editors via Omnisharp language server is really good and you can develop on any platform be it a Linux distro or M1 MacBook. .NET is no longer a Windows-first product.

lars-berger:
If you don't have the time to work on this right now, you can [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). Think we just have to change `Bus` and that `launch.json` property, then this should be good to merge.

Regarding Omnisharp, it's pretty damn impressive how well C# works in VSCode now because of it. I remember trying it out about 3 years ago and having a bad time with it (was also a C# noob though). But now, it's been working flawlessly and I haven't felt any need at all to boot up Visual Studio for my use cases.

neon-sunset:
@lars-berger Apologies, this week was a bit busy but I think I'll be able to get back with fixes either today or tomorrow.

lars-berger:
@neon-sunset Been trying out the Roslynator extension for a little bit now and I'm glad you made the suggestion. However, I definitely ran into some confusions while getting set up. Maybe you have some ideas here.

1. There's an overwhelming amount of ways to configure Roslynator (eg. via `.editorconfig`, `.roslynatorconfig`, `*.ruleset` files, and `roslynator.config`). Would you happen to have any recommendations on how to structure these configs?
2. Also tried out the the CLI, but it doesn't seem in-line with the warnings reported in VSCode. When running `roslynator analyze`, it currently reports 71 rule issues, but under the "Problems" tab, there's 200+ warnings. Have you run into this before?
3. For applying the rule fixes, were you doing it via the VSCode command palette `Omnisharp: Fix all occurrences of a code issue within solution`? This goes through the rules 1-by-1, so I'm hoping there's a faster way if the CLI continues to be troublesome.

PR #47, opened 2022-04-18
feat: add `swap_monitors` command for moving workspaces between monitors

matt-andrews:
Love your app!

I hope PR's are ok, I had a desire to swap workspaces between my two monitors so I added a command for it. I didn't see any contribution guidelines so I did my best to retain consistency. Let me know if you need me to make any adjustments.

Keep up the good work!

### DIFF
--- a/GlazeWM.Domain/DependencyInjection.cs
+++ b/GlazeWM.Domain/DependencyInjection.cs
@@ -67,6 +67,7 @@ namespace GlazeWM.Domain
       services.AddSingleton<ICommandHandler<DisplayWorkspaceCommand>, DisplayWorkspaceHandler>();
       services.AddSingleton<ICommandHandler<FocusWorkspaceCommand>, FocusWorkspaceHandler>();
       services.AddSingleton<ICommandHandler<MoveFocusedWindowToWorkspaceCommand>, MoveFocusedWindowToWorkspaceHandler>();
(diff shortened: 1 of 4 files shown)


lars-berger:
Hey, this is seriously cool. Thanks for putting this together 👍

Personally running 2 monitors and it's working great. However, my concern is that it's not as useful in different monitor setups and new commands would have to be added to fill that gap.

Say a user has 3 monitors each with its own collection of workspaces. If the monitors are initially ordered as `[monitor A, monitor B, monitor C]` and since it's currently always shifting to the following monitor, it wouldn't be possible to swap the monitors to make it `[B, C, A]` `[A, C, B]`, or `[C, B, A]`.

Instead of shifting the workspaces to the next monitor in `MonitorService.GetMonitors()`, an improvement that could be made is to make the command directional (eg. `swap monitors left`). There's a method on `MonitorService` for getting a monitor in a given direction (see [here](https://github.com/lars-berger/GlazeWM/blob/fe1bae405593c8a8094237ea953e53632ca003ff/GlazeWM.Domain/Monitors/MonitorService.cs#L90))

lars-berger:
Closing since the OP is inactive. There's also been a new command added for moving a single workspace in a given direction (`Alt+<ASDF>` in default keybindings), which fulfills a similar use case.

PR #60, opened 2022-04-21
feat: add some default window rules to sample config

scubamount:
> includes some optional commented examples in the sample config that are like that of the README.md, with the inclusion of dialog boxes from https://github.com/lars-berger/GlazeWM/issues/58

### DIFF
--- a/GlazeWM.Bootstrapper/sample-config.yaml
+++ b/GlazeWM.Bootstrapper/sample-config.yaml
@@ -141,18 +141,18 @@ keybindings:
     bindings: ["Alt+Shift+9"]
 
 #LIST OF OPTIONAL IGNORED WINDOWS FOR QUALITY-OF-LIFE
-#window_rules:
-#  - command: "ignore"
-#   match_process_name: "/explorer/"
+window_rules:
+  - command: "ignore"
+    match_process_name: "/explorer/"
 
-#  - command: "ignore"
-#    match_process_name: "/Taskmgr/"
+  - command: "ignore"
+    match_process_name: "/Taskmgr/"
 
-#  - command: "ignore"
-#   match_title: "/Save As/"
+  - command: "ignore"
+    match_title: "/Save As/"
 
-#  - command: "ignore"
-#   match_title: "/Open/"
+  - command: "ignore"
+    match_title: "/Open/"
 
-#  - command: "ignore"
-#   match_process_name: "/notepad/"
+  - command: "ignore"
+    match_process_name: "/notepad/"
--- a/GlazeWM.Bootstrapper/sample-config.yaml
+++ b/GlazeWM.Bootstrapper/sample-config.yaml
@@ -141,18 +141,18 @@ keybindings:
     bindings: ["Alt+Shift+9"]
 
 #LIST OF OPTIONAL IGNORED WINDOWS FOR QUALITY-OF-LIFE
-window_rules:
-  - command: "ignore"
-    match_process_name: "/explorer/"
+#window_rules:
+#  - command: "ignore"
+#    match_process_name: "/explorer/"
 
-  - command: "ignore"
-    match_process_name: "/Taskmgr/"
+#  - command: "ignore"
+#    match_process_name: "/Taskmgr/"
 
-  - command: "ignore"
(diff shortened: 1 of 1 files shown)


lars-berger:
Hey, thanks for the PR 👍

Window rules are quite opinionated, so it's tricky to make a judgment on what would be universally useful. It would be nice to have some simple rules added to the default config for the sake of showcasing the feature though, and I think some of the rules in this PR are good defaults.

Regarding the specific rules added in the PR:
* Task Manager - Definitely agree it should be ignored by default. Since it runs as elevated, it can't be managed unless the WM is also run as admin.
* Open/save File Explorer - Also agree. Since it's a dialog that users typically open and close quickly, it can be annoying for this to tile by default. Instead of using `ignore` though, I think it'd make more sense to float the window via `toggle floating`.
* Notepad and Explorer - Personally think this is too opinionated to add as a default rule. 

Will add a PR to your fork in just a sec with some suggested changes and we can get this merged in

lars-berger:
Well, apparently I can't PR your fork hehe. Still pretty new to some of the conventions in open source.

Here's the suggested change:
```yaml
window_rules:
  # Task Manager requires admin privileges to manage and should be ignored unless running the WM as admin.
  - command: "ignore"
    match_process_name: "Taskmgr"

  # Launches the File Explorer save/open dialogs as floating by default.
  - command: "toggle floating"
    match_title: "/^(Open|Save As)/"
    match_class_name: "#32770"
```
I think it'd be fine to not comment them out and simply leave them as default active rules.

nascentt:
I'd recommended adding Find to `match_title: "/^(Open|Save As)/"`

`match_title: "/^(Open|Save As|Find)/"`

nascentt:
I found this rule has been working well but fails when opening downloaded files with an Open File - Security Warning. I assume because it's a security window that glazewm cant access or something?

`- command: "toggle floating"
    match_title: "/(Open|Save As|Find)/"
    match_class_name: "#32770"`

would be good to float the open with dialog 
open file security warning


also a similar rule I definitely recommend for default is the copy file dialog

```
- command: "toggle floating"
    match_class_name: "OperationStatusWindow"
```

lars-berger:
@nascentt  For floating the dialogs with class `#32770`, it might make sense to just remove the title match. The class is mentioned in the docs as being ["The class for a dialog box."](https://docs.microsoft.com/en-us/windows/win32/winmsg/about-window-classes#system-classes), so I guess it'd be pretty safe to simply float it by default. And apps can actually customize the window title of the file open dialogs, so it's not super reliable to match by title.

Also, good catch with the copy dialog - that should definitely be floated.

@scubamount If you don't have the time to work on this right now, you can [allow edits from maintainers](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork). Something like this should be fine:

```yaml
window_rules:
  # Task Manager requires admin privileges to manage and should be ignored unless running the WM as admin.
  - command: "ignore"
    match_process_name: "Taskmgr"

  # Launches system dialogs as floating by default (eg. File Explorer save/open dialog).
  - command: "toggle floating"
    match_class_name: "#32770"

  # Launches file copy & move dialog as floating by default.
  - command: "toggle floating"
    match_class_name: "OperationStatusWindow"
```

lars-berger:
Actually realized there were some issues with this window rule shortly after merging:

```yaml
  # Launches system dialogs as floating by default (eg. File Explorer save/open dialog).
  - command: "toggle floating"
    match_class_name: "#32770"
```

Some windows with class `#32770` were already floating by default, so this ended up toggling them to be tiling. Added a new command `"set floating"` to deal with this, so the new rule looks like this:

```yaml
  # Launches system dialogs as floating by default (eg. File Explorer save/open dialog).
  - command: "set floating"
    match_class_name: "#32770"
```

Also, ended up moving the below rule out of the sample user config to some hidden default window rules in the code. I can't think of any case where a user would want these windows to actually be tiling, so it made more sense to not expose the rule in the user config.

```yaml
  # Launches file copy & move dialog as floating by default.
  - command: "toggle floating"
    match_class_name: "OperationStatusWindow"
```

PR #83, opened 2022-06-14
feat: Add ability to set the font weight

KiLLeRRaT:
I've added the ability to set the font weight via config.  Defaults to `Normal`.

Possible values are as per https://docs.microsoft.com/en-us/dotnet/api/system.windows.fontweights?view=windowsdesktop-6.0#remarks


KiLLeRRaT:
> Apologies for getting to this so late. This looks great, thank you! 👍

Awesome, thanks :)

PR #86, opened 2022-07-04
bug: Added code to make sure that the keys pressed matches the keybind exactly

KiLLeRRaT:
This resolves the issue where a keybind has less keys than the keys that were pressed.

E.g. when you have a keybind of `Alt+1`, but you press `Control+Alt+1`, it still fired.  It should NOT fire it, because you didn't press exactly `Alt+1`.

I had a problem where I have `Control+Alt+1` bound in another application and Glaze hijacked the keybinding, despite my config only being `Alt+1`.

Let me know if this is fine, or if there is a better way to handle this.

Cheers,

### DIFF
--- a/GlazeWM.Infrastructure/WindowsApi/KeybindingService.cs
+++ b/GlazeWM.Infrastructure/WindowsApi/KeybindingService.cs
@@ -31,6 +31,15 @@ namespace GlazeWM.Infrastructure.WindowsApi
     /// </summary>
     private readonly Dictionary<Keys, List<Keybinding>> _keybindingsByTriggerKey = new();
 
+    private readonly Keys[] _modifierKeys = new Keys[] {
+      Keys.LControlKey,
+      Keys.RControlKey,
+      Keys.LMenu,
+      Keys.RMenu,
+      Keys.LShiftKey,
+      Keys.RShiftKey
+    };
+
     public void Start()
     {
       var thread = new Thread(() => CreateKeybindingHook())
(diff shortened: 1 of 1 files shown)


KiLLeRRaT:
This had a bug with the loop, so I've changed it to 3 if statements, checking each of the modifier keys.

lars-berger:
Hey, this is a solid change - definitely agree that this is better behavior.

Here's a way to get the loop check working

```csharp
private IntPtr KeybindingHookProc(...)
{
  ...
  // Get modifier keys that aren't part of the key combination.
  var modifierKeysToReject = modifierKeys.Where(
    (modifierKey) =>
      !longestKeybinding.KeyCombination.Contains(modifierKey) &&
      !longestKeybinding.KeyCombination.Contains(GetGenericKey(modifierKey))
  );

  var hasModifierKeysToReject = modifierKeysToReject.Any((modifierKey) =>
    cachedKeyStates.ContainsKey(modifierKey)
      ? cachedKeyStates[modifierKey]
      : IsKeyDown(modifierKey)
  );

  // This makes sure that if you press Control+Alt+1, and you have a keybinding of Alt+1, that
  // it doesn't fire. Even though that all the keys satisfy it, the event has too many keys,
  // so it shouldn't fire.
  if (hasModifierKeysToReject)
    return CallNextHookEx(IntPtr.Zero, nCode, wParam, lParam);
   ...
}
 
private static Keys GetGenericKey(Keys key)
{
  return key switch
  {
    Keys.LMenu or Keys.RMenu => Keys.Alt,
    Keys.LShiftKey or Keys.RShiftKey => Keys.Shift,
    Keys.LControlKey or Keys.RControlKey => Keys.Control,
    _ => key,
  };
}
```

We can move out this check till after the `longestKeybinding` null check. This should hopefully be more performant since we'd only check whether extra modifier keys are pressed only when a keybinding actually matches.

lars-berger:
Should hopefully get this out in a new release by the end of the week, thanks again @KiLLeRRaT 🙏

PR #108, opened 2022-08-22
(RDY)feat: add support for launching user defined processes

### LINKED ISSUE #104: Feature request: shortcut for launching default terminal/editor
Hi, first of all, I'd like to say that this is the best WM I've used on Windows so far; thank you for the excellent work!

Something convenient I've seen in typical linux WMs is the ability to launch a terminal/editor with a shortcut, and I was wondering if that is something you would consider for Glaze as well? There are already many shortcuts that depend on the alt key, so it's understandable if you'd prefer to not add yet another one or two, but something like alt+t to launch your preferred terminal program would be neat for example.

I can definitely have a go at making a PR if you'd like to consider this feature.

AreamanM:
Closes #104 

This PR adds support for launching a user defined process from a keybind.

Here's an example with launching Visual Studio Code with no arguments:

image

If this looks good so far, I can go ahead with implementing logging and other bits and pieces for the new classes that were implemented.

#### Notes:
- Executing `cmd.exe /C start $ARGS` lead to an annoying problem where an erroneous cmd.exe window would stay open after the process specified launches, but executing `cmd.exe /C cmd.exe /C start /B` seemed to work.
- I know `start` was originally suggested for the implementation, but there's a weird quirk where a everytime you launch a process using `start`, a `Console Window Host`(conhost) process pops up, and even after the user specified process that was started closes, conhost stays and hogs ram. So for e.g. if you start `code` 5 times, 5 corresponding `Console Window Host` processes start and never end even if VSC itself is closed, so they constantly hog memory. 

- Executing `cmd.exe /C $ARGS` directly avoids this issue of conhost processes firing up and staying in memory and has similar semantics to `cmd.exe /C start $ARGS`.
- But since `start` is not being used, if the executable name in the config does not exist, there is no error information that is displayed, it just stays silent and does nothing. (Mabye `Process.ExitCode` might be of use here, but I am not sure because I haven't verified if it works yet).

TODO:

- [ ] logging
- [ ] documentation


lars-berger:
A pretty major difference between running `cmd.exe /C $ARGS` vs `cmd.exe /C start $ARGS` is that the first will only work for apps registered in a folder in the `PATH` environment variable (eg. `cmd.exe /C chrome` doesn't work because it's not in `$PATH`). Most apps can be launched via `start` though.

I'm trying to replicate the issue with rogue `Console Window Host` processes, but it honestly seems to be working fine when checking task mgr. Changed the args for the cmd process to `$"/C start /B {command.ProcessName} {string.Join(" ", command.Args)}",`.

https://user-images.githubusercontent.com/34844898/186089305-68200691-0b15-4b47-a70d-ee6d07f9f287.mp4

AreamanM:
I just tried it again:

image
I launched 3 VSC instances using the assigned keybind
image
there's 3 corresponding conhost processes now

Then I closed them, but the conhost processes still persisted, and launching new VSC instances launched more conhost processes that never shut down, so in a long session where the for example VSC and WT are opened and closed a lot, you could end up with 20 of these processes taking ~5 mb of ram each for no reason(they can be safely killed with no side effect on the running process or system).

Are you using Git Bash by any chance? Git Bash ships with it's own `start` executable that is in PATH, and that might have different effects compared to the cmd builtin.

lars-berger:
Huh it actually seems like the problem is with VSCode specifically.

Like when just running `start <PROCESS>` in a cmd terminal, there's a difference in how VSCode behaves vs all other apps I've tested out. If you take a look at the vid, what I'm guessing is happening is this cmd window that gets spawned when doing `start code` just doesn't get killed when called in C#.

https://user-images.githubusercontent.com/34844898/186196549-b4299c77-4be0-431d-b15e-f095e998a9bc.mp4

When launching Code via `LWin+R` dialog, there is also a cmd window that pops up, but it gets killed after a split second. Not sure how to replicate such behavior. Another thing that's confusing to me is why in the vid the new cmd windows don't show up as a `Console Window Host` in taskmgr, but they show up when launching Code via C#.

Had no idea about the different `start` in Git Bash btw. It doesn't seem to be messing with the behavior.

AreamanM:
Hmm that explains why you couldn't replicate the issue earlier, I think it's definitely worth using `start`, and this issue should be reported and handled by the VSCode team.

Although I do suspect the reason this happens is that when `code` is in $PATH, `start code` invokes a `code.cmd` file:
image

which is located in %LOCALAPPDATA%\Programs\Microsoft VS Code\bin for a single user install.

I renamed both, the shell script and the `.cmd` file, and then tried `start code`. The run dialog complained that Windows could not find `code`, which is a bit unfortunate.

The `code.cmd` file does seem to explain the weird conhost processes and `start code` launching stray cmd windows with VSC.

But, if you invoke code by directly executing `code.cmd` by typing `"%LOCALAPPDATA%\Programs\Microsoft VS Code\bin\code"`, it doesn't launch up an extra cmd window etc.

Is it worth adding in special handling in `StartProcessHandler`? Otherwise, regular programs seem to be responding well to `/C start $ARGS` as arguments to the `cmd.exe` process. I do think `start` should definitely be used since as you mentioned, programs not in $PATH can be launched easily as well.

lars-berger:
Yeah, adding special handling for VSCode might be the best choice here. It's probably better to have an ugly hack than a memory leak.

Will take a look tomorrow to see what's up with `code.cmd`

AreamanM:
Ok I'll quickly add in logic to directly launch `code.cmd` if the process name is `code`.

lars-berger:
It turns out that [`ShellExecuteEx`](https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecuteexa) can actually be called with strings like `chrome` or `code` instead of a path to an executable. This isn't super obvious from the MSDN docs, but it's made a lil more clear [here](https://docs.microsoft.com/en-us/windows/win32/shell/app-registration#finding-an-application-executable).

We can use `ShellExecuteEx` under the hood by adding `UseShellExecute = true` when starting the process. This means we can just completely avoid using `start` with cmd.exe and the process is started directly 🎉 

```csharp
public CommandResponse Handle(StartProcessCommand command)
{
  using var process = new Process();
  process.StartInfo = new ProcessStartInfo
  {
    UseShellExecute = true,
    FileName = command.ProcessName,
    Arguments = string.Join(" ", command.Args)
  };
  process.Start();

  return CommandResponse.Ok;
}
```

Other than that, I'd basically be happy to merge as is. There's some basic logging when a command is called and the docs probably need some improvements before there's a good place to slap info about this command. If there are any follow-up changes that you do want to make though, totally feel free. 

Also debating whether it makes more sense to now rename the command to `exec ...` instead of `start ...` since it's not 1:1 with cmd's `start` command anymore and `exec` is the term used in i3. LMK what you reckon and I'll let the idea simmer a bit while queuing up some more features for the next release. Either way, renaming the command doesn't have to be part of this PR.

AreamanM:
Oh that's actually very handy, thanks. I think renaming it to exec might be a good idea, and in my opinion, `exec ...`  is a little more intuitive and will also be familiar to users of i3 and other similar WMs, so my suggestion would be to go for the rename.

There's still the issue of VSC launching a `cmd.exe` window along with itself, but I think it's worth opening a separate issue there and discussing fixes there instead.

I think the PR's ready to merge now, and I'd be happy to help with docs if needed. Do feel free to squash the commits into one big commit because the commit history is a bit sporadic.

lars-berger:
Does the CMD window linger for you when launching VSC? For me it's killed shortly after running the command, just like when running `code` via Windows Run. It's a little annoying for sure, but reckon it's fine to release as is.

Good point about the error thrown when the program doesn't exist. Will go ahead and merge this, but I'll need to address that prior to release.

PR #112, opened 2022-08-31
feat: Allow change of config file location via command line args

lyze237:
Hey!

@stewart-james asked in #41 for allowing to load a different config file, so I quickly threw together this PR. 

`new FileInfo(path).FullName` is a good replacement for the URI thing as that allows relative file paths as well.

### DIFF
--- a/GlazeWM.Bootstrapper/Program.cs
+++ b/GlazeWM.Bootstrapper/Program.cs
@@ -7,9 +7,11 @@ using Microsoft.Extensions.Hosting;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Console;
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Threading;
 using System.Windows.Forms;
+using Microsoft.Extensions.Configuration;
 
 namespace GlazeWM.Bootstrapper
 {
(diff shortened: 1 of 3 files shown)


lars-berger:
Sweet, this is much appreciated ❤️ 

.NET generic host was added to the project a lil while back, so there's a more standardized way to consume CLI args. IMO this way is more flexible if we decide to add additional CLI arguments in the future (instead of `args[1]`, `args[2]` etc.)

```csharp
// Program.cs
private static void Main(string[] args)
{
  ...
  var host = CreateHost(args);
  ...
}

private static IHost CreateHost(string[] args)
{
  return Host.CreateDefaultBuilder()
    .ConfigureAppConfiguration((_, appConfig) =>
    {
      appConfig.AddCommandLine(args, new Dictionary<string, string> {
      // Map CLI argument `--config` to `UserConfigPath` configuration key.
      { "--config", "UserConfigPath" },
      });
    })
    ...
}
```

```csharp
// UserConfigService.cs
...
public string UserConfigPath =>
  _configuration.GetValue<string>("UserConfigPath") ?? _defaultUserConfigPath;
  
private readonly IConfiguration _configuration;
private readonly string _defaultUserConfigPath = Path.Combine(
  Environment.GetFolderPath(Environment.SpecialFolder.UserProfile),
  "./.glaze-wm/config.yaml"
);

public UserConfigService(IConfiguration configuration)
{
  _configuration = configuration;
}
```

lyze237:
Oh that's great, I didn't know that existed, makes everything a lot more cleaner! :)

PR #117, opened 2022-09-03
Added some missing features to the readme

lyze237:
Hey! 

I added the `--config` command line argument and the  `bind_to_monitor` thing to the readme.

Additionally, I added all the commands in the readme as well and think that's a good thing to have. (Since I think there are some undocumented ones right now, like the `start` one.)

I'm not sure if you'd also want to add a little description to every command, if so, you'd need to help me with that please!

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,23 @@ window_rules:
     match_process_name: "notepad"
 ```
 
+# Commands
+
+* layout vertical/horizontal
+* focus left/right/up/down
+* focus workspace \<number>
+* move left/right/up/down
+* move to workspace \<number>
+* resize height/width \<amount>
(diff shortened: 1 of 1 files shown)


lars-berger:
Added a couple minor comments

Will wait to merge till the next release is ready since some of these features aren't in the current release. Need ~2 more days to wrap up my config reload feature and then i'll put out a 1.7 version.

Also, moved over all my tasks to [Github projects](https://github.com/lars-berger/GlazeWM/projects) so there's a lil better transparency. If you ever want to DM me about something, hit me up at `@lberger#4570` on discord

lars-berger (changes requested):

PR #118, opened 2022-09-05
bug: Fix #115 - Workspace name missalignment

### LINKED ISSUE #115: Workspace buttons missaligned
Hey!

I have a 3 screen setup with multiple DPI scales and noticed that workspace buttons get misaligned sometimes and show partially offscreen:

3440x1440 with 100%
image

2560x1440 with 150%
image

1920x1080 with 100%
image

It also seems like that this only happens with longer workspace names and not with "number only" workspaces.

Bar config:
```yaml
bar:
  height: 30
  position: "bottom"
  opacity: 1.0
  background: "#282A36"
  foreground: "#F8F8F2"
  font_family: "Segoe UI"
  font_size: "13"
  padding: "1 6 1 6"
  components_left:
    - type: "workspaces"
      focused_workspace_background: "#666B87"
      displayed_workspace_background: "#44475a"
      default_workspace_background: "#44475a"
  components_right:
    - type: "clock"
      time_formatting: "HH:mm - ddd dd MMM"
```

lyze237:
Hey there!

This is a very quick fix which should fix #115:
image

### DIFF
--- a/GlazeWM.Bar/Components/WorkspacesComponent.xaml
+++ b/GlazeWM.Bar/Components/WorkspacesComponent.xaml
@@ -19,6 +19,7 @@
         <Button
           x:Name="WorkspaceButton"
           Content="{Binding DisplayName}"
+          VerticalAlignment="Center"
           Command="{Binding DataContext.FocusWorkspaceCommand, ElementName=_workspacesComponent}"
           CommandParameter="{Binding Name}"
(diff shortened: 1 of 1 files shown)


lyze237:
Just noticed that there's different behavior between number only and with names, so I'll look into that as well: 
image

Issue #156, opened 2022-10-28
Show error message on invalid keybindings


kuznetsovvj:
A copy of the draft from the project to link the merge request.

> For example, the following keybinding will silently throw the error Requested value '<CHARACTER>' was not found.

```
keybindings:
  - command: "focus left"
    binding: "Alt+;"
```

> When validating the user config (ie. ValidateDeserializedConfig method in EvaluateUserConfigHandler.cs), check whether the binding can be cast to Keys enum and throw FatalUserException if it's invalid.

kuznetsovvj:
I created pull request #157

kuznetsovvj:
Done

PR #157, opened 2022-10-28
feat: show error message on invalid keybindings (#156)

kuznetsovvj:


lars-berger:
This is great, thank you! And the code quality looks very solid 👍 

Will go ahead and merge this, but just know that it might take some time before a new release comes out. I need to build up some more features for the next release.

Issue #155, opened 2022-10-27
Be able to focus to a next/previous workspace according to order in `workspaces` config


azinsharaf:
what is the key binding to focus to a next/previous workspace? for instance if I am on WS2 which key binding focusses on WS1 and WS3?

avjinder:
You can press Alt+<n> to go to the nth workspace. Glazewm doesn't have any keybindings to go to the next or previous workspaces, you have to enter the workspace number yourself.

azinsharaf:
Could this be a feature request then?

lars-berger:
Sure, think this would be a good feature request 👍 

It's a pretty starter issue if someone wants to take a stab at this feature. Would need to add 2 new commands `"focus workspace next"` and `"focus workspace prev".

kuznetsovvj:
@lars-berger 

I tried to do the task and I have a question

By default in the config, workspaces are named 1-9. I assume that when calling `focus workspace next`, the workspace with number +1 should be selected (for number 9, number 1, respectively).

But, the implementation in the code doesn't prevent you from using strings instead of numbers in the workspace name. Which workspace should be selected if we have string names workspaces? Switch to the next name in lexicographic order?

lars-berger:
@kuznetsovvj Ay awesome, I appreciate you taking a look.

What you could do is get the index of the config for the currently focused workspace in `UserConfigService.WorkspaceConfigs` (to get the focused workspace see: `WorkspaceService.GetFocusedWorkspace()`). Then you could call `FocusWorkspaceCommand` with the name of the workspace at the next/prev index

lars-berger:
@kuznetsovvj since "piupiu" is focused, "miyau" should get focus (index of "piupiu" + 1). If `focus workspace prev` is called, then it'd be the one before "piupiu", so "bullet".

kuznetsovvj:
@lars-berger 

I did not add command `focus workspace next` and `focus workspace prev` to the default config because I do not know what key combination to use

kuznetsovvj:
@lars-berger maybe close this issue? feature is done

azinsharaf:
Thanks.

lars-berger:
This is now a part of the [latest release](https://github.com/lars-berger/GlazeWM/releases)! 🎉

PR #159, opened 2022-10-31
feat: add commands "focus workspace next/prev" (#155)

kuznetsovvj:
issue #155


lars-berger:
Made a slight change to this feature post-merge btw. Made it so that it iterates only through active workspaces instead.

Just wanted to let you know, hopefully you're ok with the change 👌

Issue #164, opened 2022-11-09
New bar component (text) is not in README.md


ColonelBucket8:
Under Bar Component Configuration section, there are only 2 components listed here. 
https://github.com/lars-berger/GlazeWM#bar-component-configuration
```
# Type of component to display. Currently only 2 component types exist:  "workspaces" and "clock".
type: <COMPONENT_TYPE>
```
Should include the new text component as it's released in [v1.8.0](https://github.com/lars-berger/GlazeWM/releases/tag/v1.8.0)

lars-berger:
Updated by @afermar's PR 👍

PR #165, opened 2022-11-09
Update README to add text component (#164)

afermar:
Added reference to text component in README to address #164


lars-berger:
Nice 👌

PR #167, opened 2022-11-10
fix: windows not hiding correctly after switching workspace focus

### LINKED ISSUE #163: bug: Switching to a workspace on another monitor does not hide windows from the workspace that was previously focused on that monitor
Say for example that my setup includes two monitors, each of which have two workspaces bound to them. Let's call the monitors A and B. My workspaces would be laid out like this:

Monitor | Workspaces
-|-
A|1, 3
B|2, 4

What I'm observing is that if Workspace 1 is focused on Monitor A, Workspace 2 is focused on Monitor B, and globally, I have a window in Workspace 1 focused, if I switch directly to Workspace 4 (changing the workspace on Monitor B, then), the window(s) contained in Workspace 2 will not be hidden.

If in the example I instead switched from Workspace 1 to Workspace 2 (or had a window in Workspace 2 gain focus any other way, like by clicking one with my mouse), *then* to Workspace 4, the Workspace 2 window(s) would be hidden correctly.

This behavior can be difficult to notice with only one window per workspace, but if you have multiple windows per workspace (especially on the destination workspace, Workspace 4 in the example), you usually wind up with an old, should-be-hidden window covering up a window you're trying to see.

lars-berger:
* Fixes issue where hidden workspaces are sometimes shown at the same time as the displayed workspace. 

Closes #163

### DIFF
--- a/GlazeWM.Domain/Workspaces/CommandHandlers/MoveWindowToWorkspaceHandler.cs
+++ b/GlazeWM.Domain/Workspaces/CommandHandlers/MoveWindowToWorkspaceHandler.cs
@@ -54,10 +54,18 @@ namespace GlazeWM.Domain.Workspaces.CommandHandlers
 
       var focusTarget = WindowService.GetFocusTargetAfterRemoval(windowToMove);
 
+      // Since the workspace that gets displayed is the last focused child, focus needs to be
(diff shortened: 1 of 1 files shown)


jjkinsey:
Thanks so much for the quick fix. Built master and it's running great.

PR #160, opened 2022-10-31
feat: add "focus workspace recent" command (#151)

kuznetsovvj:

PR #178, opened 2022-11-24
feat: add "focus workspace toggle" command

gtsherman:
This pull request adds the ability to toggle focus between the current workspace and the most recent workspace using a new command, `focus workspace toggle`.

These updates are modeled on #160. The functionality is similar, but `focus workspace recent` is more like an "undo" command. For example, say you focus workspace 1, then workspace 2, then workspace 3. Calling `focus workspace recent` twice would:
1. focus workspace 2
2. focus workspace 1

In contrast, `focus workspace toggle` twice would:
1. focus workspace 2
2. focus workspace 3

(and would continue to toggle until a different workspace is selected).

### DIFF
--- a/GlazeWM.Bootstrapper/sample-config.yaml
+++ b/GlazeWM.Bootstrapper/sample-config.yaml
@@ -115,9 +115,13 @@ keybindings:
   - command: "exec cmd"
     binding: "Alt+Enter"
 
-  # Focus the workspace that last had focus.
+  # Focus the workspace that had earlier focus.
   - command: "focus workspace recent"
     binding: "Alt+R"
+  
+  # Toggle focus to the most recent workspace.
+  - command: "focus workspace toggle"
+    binding: "Alt+Oemtilde"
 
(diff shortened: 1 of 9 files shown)


lars-berger:
I think it's gonna be confusing with 2 commands that are so similar. I do agree that this behavior is more intuitive though, and it could also be reused for implementing something like i3's `workspace_auto_back_and_forth` in the future.

Would you be ok with instead changing the existing `focus workspace recent` command (either in this or a separate PR)? If not, I'm fine with making the change as well and I'll credit you in the release

ericreeves:
Just tossing in my 2 cents that I agree that this is the more intuitive behavior for "recent".  It's what I expected to happen, and I actually didn't understand what was happening until I read this PR and realized the current behavior is a "workspace switch undo history", which is not something I've ever seen in a WM.

+1!

gtsherman:
I agree with you both -- I just didn't want to presume about replacing the existing `focus workspace recent`. I can work on subbing this functionality for that command. I'll issue a new PR when done!

PR #186, opened 2022-12-09
feat: binding modes

### LINKED ISSUE #181: Feature Request: binding modes
Props for creating what is probably the "KISSest" tiling WM for windows, a great experience so far.

I know this is probably not meant to be a full i3 clone, but it would be really nice to have [binding modes](https://i3wm.org/docs/userguide.html#binding_modes)

lars-berger:
* Add new command `binding mode <BINDING_MODE_NAME>`. When a binding mode is active, only the keybindings defined for that binding mode are available. Use `binding mode none` to exit the binding mode and return to the default keybindings.
* Add user config property `binding_modes`.
* Change sample config to use a binding mode for resize commands.

Closes #181


gRox167:
I am wondering if we can exit the binding mode right after we triggered an command? This will be helpful for command like changing focus, or moving windows. I tried to use 2 terms in a list but there is a error reported when I reload the config. I am wonder if the multiple commands option in binding_modes are implemented?

PR #187, opened 2022-12-10
Made "focus workspace recent" a toggle

gtsherman:
Per #178, updates the "focus workspace recent" command to toggle between workspaces (instead of undo-like functionality).


lars-berger:
Awesome, thank you 🙌 This is a lot more intuitive

Next release should come out in the next 1-2 weeks with this included.

Issue #196, opened 2022-12-20
Feature request: Calendar week in clock widget


s-v-o:
I often need the calendar week in my daily business.
It would be nice when I could configure the clock component so, that the calendar week is displayed.

In the [standard formatting](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings) I could not find any directly working possibility.

Also the [calendar week](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.calendar.getweekofyear?view=net-6.0#system-globalization-calendar-getweekofyear(system-datetime-system-globalization-calendarweekrule-system-dayofweek)) can differ based on the used [CalendarWeekRule](https://learn.microsoft.com/en-us/dotnet/api/system.globalization.calendarweekrule?view=net-6.0) and [FirstDayOfWeek](https://learn.microsoft.com/en-us/dotnet/api/microsoft.visualbasic.firstdayofweek?view=net-6.0).

lars-berger:
This is now part of the [latest release](https://github.com/lars-berger/GlazeWM/releases)! Thanks for the help with this one 👌

PR #197, opened 2022-12-20
feat(bar): calendar week in clock component

s-v-o:
Pull request to show calendar week in clock component (#196).

### DIFF
--- a/GlazeWM.Bar/Components/ClockComponentViewModel.cs
+++ b/GlazeWM.Bar/Components/ClockComponentViewModel.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Globalization;
 using System.Reactive.Linq;
 using GlazeWM.Domain.UserConfigs;
@@ -22,9 +22,9 @@ namespace GlazeWM.Bar.Components
         var additionalInfo = "";
         if (_config.WeekOfYear)
         {
-          additionalInfo = additionalInfo + CultureInfo.CurrentCulture.Calendar.GetWeekOfYear(now, _config.CalendarWeekRule, _config.FirstDayOfWeek) + " CW";
+          additionalInfo = additionalInfo + " (" + CultureInfo.CurrentCulture.Calendar.GetWeekOfYear(now, _config.CalendarWeekRule, _config.FirstDayOfWeek) + " CW" + ")";
         }
-        return (dateString + " (" + additionalInfo + ")");
+        return (dateString + additionalInfo);
       }
     }
 
--- a/GlazeWM.Bar/Components/ClockComponentViewModel.cs
+++ b/GlazeWM.Bar/Components/ClockComponentViewModel.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Globalization;
 using System.Reactive.Linq;
+using System.Text;
 using GlazeWM.Domain.UserConfigs;
 
 namespace GlazeWM.Bar.Components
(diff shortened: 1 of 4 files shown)


s-v-o:
Maybe there is a simpler or more elegant way (I'am no .NET developer), but it would be sufficient and helpful for my needs and maybe for others too.

The best approach in my opinion would be to configure it with the time_formatting string, but I hesitated to mess around with standard behavior.

lars-berger:
Are you Swedish? 👀 We'd use CW sometimes at my old workplace, but I think it's fairly niche for the rest of the world

I'm hesitant to merge this given how few people would likely use this, as well as how much extra configuration it adds for the clock component.

s-v-o:
No, I'm from Germany, but once I was in Sweden on vacation and it was wonderful.  
We use CW for rough planning and faster calendar navigation (i.e.: "is there a meeting timeslot in CW 52").

> I'm hesitant to merge this given how few people would likely use this, as well as how much extra configuration it adds for the clock component.

I understand. Do you think it would be a suitable or better approach to use the time_formatting string instead?

I could take inspiration from Javas [SimpleDateFormat (Java Platform SE 6)](https://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html) and use 'w' for calendar week (I think in that case I only want to support the ISO 8601 variant (see also [ISOWeek.GetWeekOfYear(DateTime) Methode (System.Globalization) | Microsoft Learn](https://learn.microsoft.com/de-de/dotnet/api/system.globalization.isoweek.getweekofyear?view=net-7.0#system-globalization-isoweek-getweekofyear%28system-datetime%29) or [GregorianCalendar (Java Platform SE 7 )](https://docs.oracle.com/javase/7/docs/api/java/util/GregorianCalendar.html#week_and_year)) 

Example:
'HH:mm dd.MM.yyyy (ww CW)' => 13:05 21.12.2022 (51 CW)

When different options should be supported, I have also seen this variant: [Using Week-Based Placeholders in Custom Date Formats - Tableau](https://help.tableau.com/current/pro/desktop/en-us/dates_custom_date_formats_week.htm)

Example:
'HH:mm dd.MM.yyyy (ww[Y] CW)' => 13:05 21.12.2022 (51 CW)
'HH:mm dd.MM.yyyy (ww[y] CW)' => 13:05 21.12.2022 (52 CW)

lars-berger:
True, it seems like a lot of the date-time formatters for other languages have support for the week number. The time formatter for C++ also has support for it, using `U`, `V` or `W` via [`std::put_time`](https://en.cppreference.com/w/cpp/io/manip/put_time).

If you're up for it, I think it'd be fine to add week number to the `time_formatting` config string 👍 One concern I had was that it makes it more unclear which symbols are available. But the .NET time formatting docs are very developer oriented, so it'd perhaps be more user friendly anyways to list out the available symbols in the glazewm docs sometime in the future.

Also leaning towards either `w` or `W` for denoting week number, since that seems like it might be most consistent with the naming of the other symbols in .NET's time formatter

s-v-o:
I hope this solution is better.
At least it would be the solution I would have expected from the standard library.

For documentation I have added some comments in the `sample-config.yaml` and in the `README.md`.

Merry christmas and a happy new year. | God jul och gott nytt år.

lars-berger:
Thanks for the great documentation - this looks solid. Hope you had a great christmas + new year👌

PR #202, opened 2023-01-07
feat: add custom text to tiling direction component

jovark:
Hi, 
I recently started using the tiling direction component and it's very useful. My only problem with it was that I couldn't set a custom text for vertical/horizontal mode. So I've added this feature myself. I hope that this will be useful to others as well. 

If there are any problems with this or any changes that need to be made let me know. 

Thank you for making this and keep up the good work!
Joonatan


lars-berger:
This looks great as is, thank you @jovark 🙌

PR #203, opened 2023-01-09
feat: show floating on top

lars-berger:
* Be able to show floating windows as always on top via new config option `general.show_floating_on_top: true|false`.


davidc360:
Bug: This was turned on by default for me until I specified "false"

PR #206, opened 2023-01-15
fix(bar): tiling direction component updates correctly

notblam:


GlazeWM.Bar/Components/TilingDirectionComponentViewModel.cs
```diff
@@ -40,6 +40,10 @@ public class TilingDirectionComponentViewModel : ComponentViewModel
       _bus.Events.OfType<LayoutChangedEvent>().Subscribe(_ =>
         _dispatcher.Invoke(() => OnPropertyChanged(nameof(TilingDirectionString)))
       );
+
+      _bus.Events.OfType<FocusChangedEvent>().Subscribe(_ =>
+        _dispatcher.Invoke(() => OnPropertyChanged(nameof(TilingDirectionString)))
+      );
```
lars-berger, line 46: instead of 2 separate listeners, could merge into 1:

```csharp
      _bus.Events.Where(
        (@event) => @event is LayoutChangedEvent or FocusChangedEvent
      ).Subscribe((_) =>
        _dispatcher.Invoke(() => OnPropertyChanged(nameof(TilingDirectionString)))
      );
```

PR #208, opened 2023-01-18
feat/bar-config-per-monitor

notblam:


GlazeWM.Bootstrapper/sample-config.yaml
```diff
@@ -7,40 +7,41 @@ general:
...
-      margin: "0 4px 0 0"
-      padding: "0 8px"
-    - type: "clock"
-      # supported time format specifier:
-      # https://learn.microsoft.com/en-us/dotnet/standard/base-types/custom-date-and-time-format-strings
-      # "w"   Calendarweek (ISO 8601 week) 1..53
-      # "ww"  CalendarWeek (ISO 8601 week) 01..53
-      time_formatting: "hh:mm tt  ddd MMM d"
-      margin: "0 0 0 10px"
+bars:
```
lars-berger, line 10: Single bar config via `bar`  seems better as default

GlazeWM.Domain/UserConfigs/BarConfig.cs
```diff
@@ -4,6 +4,8 @@ namespace GlazeWM.Domain.UserConfigs
 {
   public class BarConfig
   {
+    public int BindToMonitor { get; set; } = 1;
```
lars-berger, line 7: Reckon we should remove the default and instead validate that this is defined in the handler for setting config

GlazeWM.Domain/UserConfigs/UserConfigService.cs
```diff
@@ -16,7 +17,7 @@ public class UserConfigService
 
     public GapsConfig GapsConfig => UserConfig.Gaps;
     public GeneralConfig GeneralConfig => UserConfig.General;
-    public BarConfig BarConfig => UserConfig.Bar;
+    public List<BarConfig> BarsConfig => UserConfig.Bars;
     public List<WorkspaceConfig> WorkspaceConfigs => UserConfig.Workspaces;
```
lars-berger, line 21: Breaks for when user defines a single bar. Ref keybindings config for inspo on how yhis can be solved

GlazeWM.Domain/UserConfigs/UserConfigService.cs
```diff
@@ -99,6 +100,23 @@ public BindingMode GetBindingModeByName(string bindingModeName)
...
+    {
+      var result = Regex.Match(monitor.DeviceName, @"\d+$", RegexOptions.RightToLeft);
+      if (result.Success)
+      {
+        int monitorIndex = int.Parse(result.Value);
+        BarConfig config = UserConfig.Bars.Find(x => x.BindToMonitor == monitorIndex);
+        // If no config was found, default the bar to the main monitor's config
+        if (config == null)
+        {
+          return UserConfig.Bars.Find(x => x.BindToMonitor == 1);
```
lars-berger, line 113: Use bar config at index 0 instead. Otherwise could be null

PR #217, opened 2023-01-20
feat: add custom X & Y offset for bar position

notblam:


notblam:
Fixed the conflicts from the feat/bar-config-per-monitor merge

GlazeWM.Bootstrapper/sample-config.yaml
```diff
@@ -8,6 +8,8 @@ general:
   cursor_follows_focus: true
 
 bar:
+  gap: false
+  corner_radius: "0"
```
lars-berger, line 12: rename to `border_radius`

GlazeWM.Domain/UserConfigs/BarConfig.cs
```diff
@@ -4,6 +4,9 @@ namespace GlazeWM.Domain.UserConfigs
 {
   public class BarConfig
   {
+    public bool Gap { get; set; }
```
lars-berger, line 7: to make the offset more customizable, we could add properties for the x/y

```yaml
bar:
  offset_x: "5px"
  offset_y: "10px"
```

GlazeWM.Domain/Workspaces/Workspace.cs
```diff
@@ -47,11 +47,13 @@ private int _yOffset
     }
 
     private int _outerGap => _userConfigService.GapsConfig.OuterGap;
-
-    public override int Height => Parent.Height - (_outerGap * 2) - _logicalBarHeight;
+    /// <summary>
+    /// If gap is enabled, correct Height and Y to make the bar floating
+    /// </summary>
+    public override int Height => _userConfigService.GetBarConfigForMonitor(Parent as Monitor).Gap ? Parent.Height - (_outerGap * 3) - _logicalBarHeight : Parent.Height - (_outerGap * 2) - _logicalBarHeight;
```
lars-berger, line 53: consider adding methods like `GetHeight()` that the getter calls to avoid long lines

GlazeWM.Bar/MainWindow.xaml
```diff
@@ -80,4 +81,4 @@
       </ItemsControl>
     </Grid>
   </Border>
-</Window>
+</Window>
```
lars-berger, line 84: add line for end of file

GlazeWM.Domain/UserConfigs/BarConfig.cs
```diff
@@ -4,6 +4,12 @@ namespace GlazeWM.Domain.UserConfigs
 {
   public class BarConfig
   {
+    public int OffsetX { get; set; }
+
+    public int OffsetY { get; set; }
+
+    public string BorderRadius { get; set; } = "7px";
+
```
lars-berger, line 12: Default to "0px"

GlazeWM.Bootstrapper/sample-config.yaml
```diff
@@ -8,6 +8,9 @@ general:
   cursor_follows_focus: true
 
 bar:
+  offset_x: 0
```
lars-berger, line 11: Change `offset_x` and `offset_y` so that they are strings that can be specified with/without pixels. Use `UnitsHelper` to trim the units and get an int

Would allow us to change this in the future to allow both px and % values without a breaking change

notblam: good call

PR #218, opened 2023-01-20
feat: add customizable separator between bar components

AreamanM:


GlazeWM.Domain/UserConfigs/BarConfig.cs
```diff
@@ -40,4 +42,11 @@ public class BarConfig
 
     public List<BarComponentConfig> ComponentsRight { get; set; } = new List<BarComponentConfig>();
   }
+
+  public class BarComponentSeparatorConfig
+  {
+    public string Left { get; set; }
+    public string Centre { get; set; }
+    public string Right { get; set; } = "|";
```
lars-berger, line 50: reckon the majority aren't too fussed about having different labels for the different sides (or just have 1 component on the left/center). so perhaps it'd be better to have a property `label` for the default, and optionally users can specify labels for specific sides

```csharp
    public string Label { get; set; } = "|"
    public string LabelLeft { get; set; }
    public string LabelCenter { get; set; }
    public string LabelRight { get; set; }
    
    // in the xaml.cs
    _barConfig.ComponentSeparators.LabelLeft ?? _barConfig.ComponentSeparators.Label
```

lars-berger: actually empty string might be a better default than `"|"`. for example, expected behavior if a user deletes the `padding` 
 property on the bar that the padding gets removed (even though it would probably look better if we set a default)

GlazeWM.Domain/UserConfigs/BarConfig.cs
```diff
@@ -40,4 +42,12 @@ public class BarConfig
 
     public List<BarComponentConfig> ComponentsRight { get; set; } = new List<BarComponentConfig>();
   }
+
+  public class BarComponentSeparatorConfig
```
lars-berger, line 46: create separate file for class

GlazeWM.Bootstrapper/sample-config.yaml
```diff
@@ -16,6 +16,8 @@ bar:
   font_family: "Segoe UI"
   font_size: "13px"
   padding: "4px 6px"
+  component_separators:
```
lars-berger, line 19: rename to singular `component_separator`

GlazeWM.Domain/UserConfigs/BarConfig.cs
```diff
@@ -40,4 +42,12 @@ public class BarConfig
...
+  {
+    public string Default { get; set; }
```
lars-berger, line 48: even though WPF doesn't error when `null` is used for a text block, think it'd be good if we explicitly set the default label to `""`

lars-berger, line 48: issue re. naming of labels - proposing we do `label`, `label_left`, etc. we can talk more on discord

GlazeWM.Bar/BarViewModel.cs
```diff
@@ -25,20 +26,51 @@ public class BarViewModel : ViewModelBase
...
-      CreateComponentViewModels(BarConfig.ComponentsCenter);
+      InsertComponentSeparator(
+        CreateComponentViewModels(BarConfig.ComponentsRight),
+         _componentSeparatorCenter);
 
     public List<ComponentViewModel> ComponentsRight =>
-      CreateComponentViewModels(BarConfig.ComponentsRight);
+      InsertComponentSeparator(
+        CreateComponentViewModels(BarConfig.ComponentsRight),
+         _componentSeparatorRight);
```
notblam, line 66: indent _componentSeperator 2 spaces instead of 1

GlazeWM.Bar/BarViewModel.cs
```diff
@@ -25,20 +26,51 @@ public class BarViewModel : ViewModelBase
...
-      CreateComponentViewModels(BarConfig.ComponentsLeft);
+      InsertComponentSeparator(
+        CreateComponentViewModels(BarConfig.ComponentsLeft),
+        _componentSeparatorLeft);
 
     public List<ComponentViewModel> ComponentsCenter =>
-      CreateComponentViewModels(BarConfig.ComponentsCenter);
+      InsertComponentSeparator(
+        CreateComponentViewModels(BarConfig.ComponentsRight),
+         _componentSeparatorCenter);
```
notblam, line 61: indent _componentSeperator 2 spaces instead of 1

PR #222, opened 2023-01-23
chore: add GitHub CI for linting

notblam:


.github/workflows/push-linter.yml
```diff
@@ -0,0 +1,23 @@
...
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Run Super Linter
+        uses: github/super-linter@v4
+        env:
+          VALIDATE_ALL_CODEBASE: false
+          DEFAULT_BRANCH: master
+          GITHUB_TOKEN: $({ secrets.GITHUB_TOKEN })
+          VALIDATE_JSCPD: false
```
lars-berger, line 21: > Once you set one of the linters to `true`, all the other linters will not run.

perhaps it'd be clearer if we explicitly run the rules we need - think it would just be `VALIDATE_CSHARP`, `VALIDATE_YAML`, `VALIDATE_XML`, and `VALIDATE_EDITORCONFIG`

notblam: I agree that would be more clear, although I'm not sure how to actually do that.  I think it runs everything TRUE by default.  So rather than setting these 4 rules to be explicitly true, we would have to set everything else to be explicitly false.

notblam: Good idea

Issue #88, opened 2022-07-19
Tray/Task Support on Bar


M1ch431:
Title 🙂

Maybe this repo could be useful in implementing this? -

[https://github.com/cairoshell/ManagedShell](https://github.com/cairoshell/ManagedShell)

lars-berger:
Hey, I really appreciate the link!

Thought that adding tray support to the bar would be completely infeasible, but this actually makes it doable. From a quick glance, the documentation also looks solid. Adding some sort of integration with `ManagedShell` would be a great addition down the line 👍

lzhecz:
I found on github program that uses that library to implement dock on windows. I use it as temporary solution as a floating tray over glazewm bar. Works for me. 

https://github.com/ze0nni/ProtoDock


A little bit changed settings and skin and how it looks on my desktop:

![GlazeWM_x64_1 9](https://user-images.githubusercontent.com/34045305/205294476-c50b9cdc-3a62-4a02-b0c4-ae45e6c60733.png)

Issue #200, opened 2022-12-29
Windows taskbar


snoopiedoom:
Hi, first of all, thanks for all the great work you've done and i really appreciate the fact someone finally made a normal wm, acting like a wm, working like a wm.

Main issue I'm facing at the moment would be the windows taskbar and I'd like some opinion and help on this matter.

For starters, I couldn't handle the gap between windows > 5 and seeing both the polybar(?) and win taskbar, so I set the windows taskbar to auto hide and it solves all of my problems, good enough. if I need something, I've the alt+space quick launch from powertoys, though it sucks ass due to slowness, but will do for now. 

Heres what I wanna do, I wanna disable the windows taskbar and get that notification center functionality up on the taskbar (it has its uses) buuuut, skimming through the docs, turns out the display manager is baked into the file explorer and desktop, and the taskbar is part of explorer. Uuuuugh...  So for now, I'd be content with glaze config having an option to remove/hide the windows taskbar, who cares about the notification center, I just don't want to watch the taskbar go through its 10 second animation every time my mouse accidentally hits the bottom of the screen. 😡 

[Here](https://superuser.com/questions/1169114/windows-10-professional-disable-taskbar) they mention replacing the shell which seems interesting but I've yet to test how much that impacts the expected windows behaviour. 
(At work, we are heavily intertwined with Microsoft)

I found [nircmd](http://www.nirsoft.net/vc/) on [this](https://superuser.com/questions/219605/how-to-completely-disable-the-windows-taskbar) post here, and the command `nircmd.exe win trans class Shell_TrayWnd 256` does make it feel like there is no taskbar, with so far no issues, but I've yet to test how he did it, I presume shell_traywnd is window, looking at it [here](https://learn.microsoft.com/en-us/dotnet/api/system.windows.window?view=windowsdesktop-7.0) and [here](https://learn.microsoft.com/en-us/windows/win32/winmsg/window-classes) makes it seem an easy job, so I guess next stp would be for me to make a fork? I don't really want to use `nircmd` as it is.

Thoughts?

P.S.
I will not give up on the dream of taking the win key from windows and turning it into the mod key.

ParasiteDelta:
[...]

PR #235, opened 2023-02-01
Feat/system tray icons component

notblam:
`Alt + LClick` to pin/unpin a system tray icon


GlazeWM.Bar/Components/SystemTrayComponentViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+
+      if (shellManager == null)
+      {
+        // Initialize the default configuration.
+        var c = ShellManager.DefaultShellConfig;
+        c.EnableTrayService = true;
+
+        // Initialize the shell manager.
+        shellManager = new ShellManager(c);
+      }
```
lars-berger, line 35: this if-block can be removed with:

```csharp
private ShellManager _shellManager = new ShellManager(ShellManager.DefaultShellConfig);
```

(`EnableTrayService` seems to be enabled by default, so no need to customize the config`

GlazeWM.Bootstrapper/GlazeWM.Bootstrapper.csproj
```diff
@@ -24,6 +24,7 @@
     <EmbeddedResource Include="icon.ico">
       <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
     </EmbeddedResource>
+    <Resource Include="Fonts\MaterialIcons-Regular.ttf" />
```
lars-berger, line 27: Rename `Fonts` dir -> `Resources` and move `icon.ico` and `sample-config.yaml` over to it

GlazeWM.Domain/UserConfigs/SystemTrayComponentConfig.cs
```diff
@@ -0,0 +1,23 @@
...
+  {
+    /// <summary>
+    /// Expands to show both pinned and unppined icons.
+    /// </summary>
+    public string ExpandText { get; set; } = "";
+
+    /// <summary>
+    /// Collapse to show only pinned icons.
+    /// </summary>
+    public string CollapseText { get; set; } = "";
```
lars-berger, line 13: rename `ExpandText` -> `LabelExpandIcon`
rename `CollapseText` -> `LabelCollapseIcon`

GlazeWM.Bar/Components/SystemTrayComponentViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+using ManagedShell.WindowsTray;
+
+namespace GlazeWM.Bar.Components
+{
+  public class SystemTrayComponentViewModel : ComponentViewModel
+  {
+    private readonly SystemTrayComponentConfig _systemTrayComponentConfig;
+    public string ExpandCollapseText { get; set; } = string.Empty;
+    public ICommand ToggleShowAllIconsCommand => new RelayCommand(ToggleShowAllIcons);
+    private bool isExpanded = true;
```
lars-berger, line 15: prefix private members with `_` eg `_isExpanded`. also add `{ get; set; }`

GlazeWM.Bar/Components/SystemTrayComponentViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+    private static NotificationArea notificationArea;
+    private static ShellManager shellManager;
```
lars-berger, line 17: afaik these don't need to be static

GlazeWM.Bar/Components/SystemTrayComponent.xaml
```diff
@@ -0,0 +1,56 @@
+<UserControl
```
lars-berger, line 1: xml indentation looks off, is it 2 spaces?

GlazeWM.Bar/Components/SystemTrayComponentViewModel.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        // Initialize the shell manager.
+        shellManager = new ShellManager(c);
+      }
+
+      notificationArea = shellManager.NotificationArea;
+      UnpinnedNotifyIcons = notificationArea.UnpinnedIcons;
+      PinnedNotifyIcons = notificationArea.PinnedIcons;
+      OnPropertyChanged(nameof(ExpandCollapseText));
+      OnPropertyChanged(nameof(UnpinnedNotifyIcons));
```
lars-berger, line 41: change these properties to automatically call `OnPropertyChanged()` whenever they are set

```csharp
    public ICollectionView _unpinnedNotifyIcons = new();
    public ICollectionView UnpinnedNotifyIcons
    {
      get => _unpinnedNotifyIcons;
      set
      {
        _unpinnedNotifyIcons = value;
        OnPropertyChanged(nameof(UnpinnedNotifyIcons));
      }
    }
```

GlazeWM.Bar/Components/NotifyIcon.xaml.cs
```diff
@@ -0,0 +1,107 @@
...
+    private ManagedShell.WindowsTray.NotifyIcon TrayIcon;
+
+    public NotifyIcon()
+    {
+      InitializeComponent();
+    }
+
+    private void NotifyIcon_OnLoaded(object sender, RoutedEventArgs e)
+    {
+      if (!isLoaded)
```
lars-berger, line 24: if my wpf knowledge is correct, this should instead have a view model (ie. `NotifyIconViewModel`). that way we don't have to grab `TrayIcon` from the `DataContext` and constantly check `isLoaded`

in `SystemTrayComponent`, we'd need a list of `NotifyIconViewModel`s and then bind that in the xaml

```xml
        <ItemsControl
            x:Name="NotifyIconsUnpinned"
            Focusable="False"
            ItemsSource="{Binding Path=UnpinnedNotifyIcons, Mode=OneWay}"> // change this to a binding to the view model list. ref `ComponentsLeft` etc for how to do this
```

GlazeWM.Bar/Components/SystemTrayComponentViewModel.cs
notblam: So without being static, the library would instantiate each time you make a bar that contains this component.  So if you ran 3 monitors, this would instantiate ShellManager 3 times.  Making it static ensures the libarary is only initializes once.

lars-berger: ah alright, that makes sense. then it'd be better imo to add shell manager to `BarViewModel` and pass it down to the ctor of `SystemTrayComponentViewModel` in the `CreateComponentViewModels` method

notblam: I was thinking of doing this way originally so that's cool with me.  I'll fix that up.

GlazeWM.Bar/Components/NotifyIcon.xaml.cs
notblam: Thanks for pointing this out.  I've been trying to read up on MVVM but it hasn't totally clicked for me yet.

GlazeWM.Bar/Components/SystemTrayComponentViewModel.cs
notblam: So upon further testing, it seems that BarViewModel is instantiated once for each bar.  I would still need to do something like mark it as static or have some checks to ensure it only loads once.  Is there something I'm missing here?

notblam:
Closes #88 #200

notblam:
closes #88

PR #238, opened 2023-02-02
Move and resize floating windows

Pamesta:


GlazeWM.Domain/Windows/CommandHandlers/ResizeWindowHandler.cs
```diff
@@ -119,5 +125,37 @@ public CommandResponse Handle(ResizeWindowCommand command)
 
       return 1.0 / resizableLength;
     }
+    private void ResizeFloatingWindow(Window windowToResize, ResizeDimension dimensionToResize, string resizeAmount)
+    {
+      int MIN_WIDTH = (int)(windowToResize.Parent.Width * 0.30);
```
lars-berger, line 130: perhaps a hardcoded pixel amount for min width/height instead (eg. 100 pixels)?

GlazeWM.Domain/Windows/CommandHandlers/ResizeWindowHandler.cs
```diff
@@ -29,6 +30,11 @@ public CommandResponse Handle(ResizeWindowCommand command)
       var resizeAmount = command.ResizeAmount;
       var windowToResize = command.WindowToResize;
 
+      if (windowToResize is FloatingWindow)
+      {
+        ResizeFloatingWindow(windowToResize, dimensionToResize, resizeAmount);
+        return CommandResponse.Ok;
+      }
```
lars-berger, line 37: think it'd be good to have a `ResizeTilingWindow` method

GlazeWM.Domain/UserConfigs/GeneralConfig.cs
```diff
@@ -11,5 +11,9 @@ public class GeneralConfig
     /// TODO: Not officially released because implementation is buggy. Use at own risk.
     /// </summary>
     public bool CursorFollowsFocus { get; set; }
+    /// <summary>
+    /// Amount by which to move floating windows
+    /// </summary>
+    public int FloatingWindowMoveAmount { get; set; } = 20;
```
lars-berger, line 17: would be good to change this to a string instead, so users could use either `20` or `20px` (then use `UnitsHelper.TrimUnits()` to get an int). also make it easier to allow % values if we add that in the future

GlazeWM.Domain/Windows/CommandHandlers/MoveWindowHandler.cs
```diff
@@ -216,5 +232,37 @@ private void ChangeWorkspaceLayout(Window windowToMove, Direction direction)
...
+          break;
+
+        case Direction.Up:
+          y -= amount;
+          break;
+
+        case Direction.Down:
+          y += amount;
+          break;
+      }
```
lars-berger, line 260: can be simplified with
```csharp
      _ = direction switch
      {
        Direction.Left => x -= amount,
        Direction.Right => x += amount,
        Direction.Up => y -= amount,
        Direction.Down => y += amount,
      };
```

.vscode/launch.json
```diff
@@ -8,7 +8,8 @@
       "name": ".NET Core Launch (console)",
       "type": "coreclr",
       "request": "launch",
-      "preLaunchTask": "build",
+      "enableStepFiltering": false,
```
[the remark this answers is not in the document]
Pamesta: sorry, modified it for myself and forgot to restore before pushing

Pamesta:
Thank you, going to clean up and edit the code based on the comments

PR #239, opened 2023-02-02
Update README to add instructions for installing via Winget

afermar:
Also added `tiling direction toggle` to the command list


notblam:
How does Winget handle updates?  Like if I install 1.11 via this method and 1.12 comes out, will it automatically update?

afermar:
Yes. In the case of portable packages, Winget overwrites the previous file with the new one.

PR #241, opened 2023-02-04
feat(infra): abstract over WASAPI COM interfaces

AreamanM:


lars-berger:
I took a quick look over at cswin32 to see if it could help us replace the `WindowsApiService` stuff. There were a few things that weren't all that great:
* We'd have to switch over a lot of types. Eg. window handles are no longer `IntPtr`, instead they're `winmdroot.Foundation.HWND`. Afaik we'd have to literally switch over every `IntPtr` handle to their corresponding cswin32 type.
image

* Although the generated comments for functions are better than what we've currently got, Vanara's seem to be way better. 
* It's easier to switch to Vanara vs cswin32 (no `NativeMethods.txt` and types should be consistent). cswin32 is also more likely to change their APIs as it matures.

Vanara's wrapper over the `CoreAudio` apis was released literally yesterday, so it's pretty perfect timing for us. I'm personally in favour of Vanara vs cswin32 or dotnet/pinvoke (similar to Vanara, but doesn't have `CoreAudio` wrappers).

If we do go ahead and switch, would `SystemVolumeInformation.cs` be simplified any further, or would it only affect `WindowsApiService`?

AreamanM:
I think Vanara is definitely the way to go, I'll merge with [refactor/replace-windows-api-pinvokes-with-vanara](https://github.com/lars-berger/GlazeWM/tree/refactor/replace-windows-api-pinvokes-with-vanara) and migrate as much of the WASAPI stuff as I can, and as for your question about SystemVolumeInformation.cs, it won't be simplified further unfortunately.

Is the structure of SystemVolumeInformation.cs fine? I'll move some classes out of the file and make the event handler part more clear, but other than that, are there any general structure changes you would suggest?

notblam:
Another big positive of going with Vanara is the code coverage is really good.  You can find the unit test for any function your interested in and get a working example with all the structs and params that are needed.  Saved me a lot of time working with the networking APIs.

Issue #145, opened 2022-10-04
Move cursor to focused workspace/window


inql:
@lars-berger thank you so much for creating such nice working i3ish WM for windows.
Is there any option to add possibility of moving mouse when changing focus? it could be very handy when working on multiple monitors :)

lars-berger:
That sounds like a good idea 👍

Haven't messed around with this feature in i3 before - it's enabled via `mouse_warping output` right? Does the cursor warp when moving the focused window via the directional move commands (eg. `Alt+H` across onto another monitor)?

akkiirah:
I've just built the latest commit and this change is really awesome.
However it would be also nice if it was possible to focus on a window when moving the mouse on one, instead of only focusing the mouse to a window if you get what I mean.

notblam:
> I've just built the latest commit and this change is really awesome. However it would be also nice if it was possible to focus on a window when moving the mouse on one, instead of only focusing the mouse to a window if you get what I mean.

Thanks for the feedback.  We still have some bugs to work out with the cursor_follows_focus feature.  I think the feature your referring to is focus_follows_cursor, which I had began working on before realizing this is a standard windows feature you can enable in the control panel.

The stock implementation works well, except for In Windows 11 there is a 500ms delay by default before a window is focused.  You can workaround this with a registry hack.
https://www.tenforums.com/tutorials/113268-change-time-activate-window-hovering-over-mouse-windows.html

akkiirah:
Thank you very much, I didn't know and expect this feature being already built into windows.
It seems working well enough :)

notblam:
> Thank you very much, I didn't know and expect this feature being already built into windows. It seems working well enough :)

So now that I've used it for a little while, it seems that it has it's bugs.  This one for example prevents me from doing any right click actions on the task bar.  
https://gyazo.com/79251640c8a0d3d11678ddeadb26d4c8

Because of this I'm reconsidering building focus_follows_cursor into GWM.  I'd like to create a hotkey and a bar component so that i could toggle this behavior on and off.  I think that might work nicely.

akkiirah:
I wasn't able to reproduce this issue just like in the clip because I observed that my Taskbar (also win11) and its menues have some sort of invisible padding which seems to prevent the cursor switching focus.
But integrating focus_follows_cursor would also be just way cleaner than telling the users "enable it in ease of use and change time via registry" imo.

michaelowens:
[...]

PR #245, opened 2023-02-11
feat: focus follows cursor

notblam:


GlazeWM.Infrastructure/WindowsApi/WindowsApiService.cs
```diff
@@ -355,12 +361,76 @@ public struct LowLevelKeyboardInputEvent
...
+      WM_RBUTTONDOWN = 0x0204,
+      WM_RBUTTONUP = 0x0205,
+      WM_KEYDOWN = 0x100,
+      WM_KEYUP = 0x101,
+      WH_KEYBOARD_LL = 13,
+      WH_MOUSE_LL = 14,
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    public struct LowLevelMouseInputEvent
```
lars-berger, line 378: the idea of having our own event that's separate from `LowLevelMouseInputEventDetails` is good 👍 

imo it should be moved out of `WindowsApiService.cs` and to its own file. change it to inherit from `Event` so that it's possible to emit on the bus and only add the properties the properties from the `LowLevelMouseInputEventDetails` struct that we actually need

```csharp
  public class MouseMoveEvent : Event
  {
    /// <summary>
    /// (X,Y) location of mouse with 0,0 being the top-left corner of the main monitor
    /// </summary>
    public Point Point { get; }

    /// <summary>
    /// Whether left-click is currently pressed.
    /// </summary>
    public bool IsMouseDown { get; }

    /// <summary>
    /// The time stamp for this message, equivalent to what `GetMessageTime` would
    /// return.
    /// </summary>
    public int TimeStamp { get; }

    public MouseMoveEvent(Point point, bool isMouseDown, int timeStamp)
    {
      Point = point;
      IsMouseDown = isMouseDown;
      TimeStamp = timeStamp;
    }
  }
```

we could add a property for `IsMouseDown` so that we keep the lower-level abstractions within `Infrastructure`

GlazeWM.Bootstrapper/Startup.cs
```diff
@@ -78,6 +86,58 @@ public void Run()
         _systemTrayIcon = new SystemTrayIcon(systemTrayIconConfig);
         _systemTrayIcon.Show();
 
+        var focusedWindows = new List<IntPtr>();
+        var isMouseDown = false;
+        if (_userConfigService.GeneralConfig.FocusFollowsCursor)
+          MouseEvents.MouseMoves.Subscribe((@event) =>
```
lars-berger, line 92: we should move most of this stuff out of `Startup` and to its own command

this also makes it easy to turn this into async (otherwise mouse movements are blocked until this command is fully ran)

```csharp
        if (_userConfigService.GeneralConfig.FocusFollowsCursor)
        // todo: avoid invoking if `IsMouseDown` is true
          MouseEvents.MouseMoves.Subscribe((@event) => _bus.InvokeAsync(new FocusContainerUnderCursorCommand(@event.Point))
 ```

lars-berger: it'd also be nice to throttle this

https://stackoverflow.com/a/3211545/9139929

it looks like `.Sample()` is the rx operator we should use. so i think this would work: `MouseEvents.Sample(TimeSpan.FromMilliseconds(50))`

notblam:
closes #145

PR #256, opened 2023-02-16
docs: Added "set tiling" command to readme

maxle5:
Reference:
https://github.com/lars-berger/GlazeWM/blob/master/GlazeWM.Domain/UserConfigs/CommandParsingService.cs#L224


lars-berger:
Thank you :ok_man:

PR #260, opened 2023-02-18
feat: add option to disable bar windows

Zinvoke:


GlazeWM.Domain/UserConfigs/GeneralConfig.cs
```diff
@@ -15,5 +15,9 @@ public class GeneralConfig
     /// Amount by which to move floating windows
     /// </summary>
     public string FloatingWindowMoveAmount { get; set; } = "5%";
+    /// <summary>
+    /// Whether to show the bar windows.
+    /// </summary>
+    public bool ShowBarWindows { get; set; } = true;
```
lars-berger, line 21: a slightly more flexible approach is if we do a property `BarConfig.Enabled` instead. that way a user could choose to have the bar enabled only on specific monitors (in combination with the multi bar config feature)

Zinvoke: Would we just iterate over each bar config to check with this approach?

Zinvoke: There each bar can individually be enabled/disabled individually

PR #268, opened 2023-02-23
feat: support absolute resize values

maxle5:
### Objective:
The resize command currently supports increasing/decreasing a window by a a certain value, however it does not currently support setting a window height/width to a specific value (e.g. set width to 25%, or set height to 500px)

### Proposal:
* If the resize command's `resizeAmount` is prefixed with "+" or "-" ([as shown in the sample config](https://github.com/lars-berger/GlazeWM/blob/master/GlazeWM.Bootstrapper/Resources/sample-config.yaml#L76)), the resize command behaves exactly as it does today (to increase/decrease the size of a window by the amount specified)
```
- command: "resize width +25%" // window grows by 25%
```

* If the `resizeAmount` is NOT prefixed with "+" or "-", we instead set the window to the size specified
```
- command: "resize width 25%" // window is resized to fill 25% of parent container 
```


lars-berger:
Thank you for this 👍 The code looks squeaky clean

However, one concern is we'd essentially have 2 distinct commands disguised as one. IMO it'd be better to just move this stuff to its own command eg. `set height 400px`

maxle5:
@lars-berger, Ok I can do that! Do we have any concerns with `set <height | width > 400px` being similar to the existing `set <floating | tiling | minimized | maximized>` command?

maxle5:
@lars-berger, should be ready to be reviewed again, let me know if there is anything else you'd like me to change!

PR #270, opened 2023-02-23
Feat/system stats component

notblam:


GlazeWM.Bar/Components/SystemStatsComponentViewModel.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+      var x = cpuCounter.NextValue();
+      double total = new Microsoft.VisualBasic.Devices.ComputerInfo().TotalPhysicalMemory;
+      var used = 1024.0 * 1024.0 * ramCounter.NextValue();
+      var ramUsage = 100.0 * (total - used) / total;
+      var gpuCounters = GetGPUCounters();
+      var gpuUsage = GetGPUUsage(gpuCounters);
+      return _config.LabelCPU + x.ToString("0.") + "%  " + _config.LabelRAM + ramUsage.ToString("0.") + "%  " + _config.LabelGPU + gpuUsage.ToString("0.") + "%";
+    }
+    private PerformanceCounter cpuCounter = new PerformanceCounter("Processor Information", "% Processor Utility", "_Total");
```
lars-berger, line 24: a lot of this could be abstracted away somewhere in `Infrastructure` eg a new `SystemStatsService` with `GetCpuUsage` methods etc

GlazeWM.Domain/UserConfigs/RAMStatsComponentConfig.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// <summary>
+    /// Icon to represent RAM usage.
+    /// </summary>
+    public string LabelRAM { get; set; } = "";
+    /// <summary>
+    /// Sets default icon font if one isn't specified.
+    /// </summary>
+    public RAMStatsComponentConfig()
+    {
+      Padding = "0px 5px";
```
lars-berger, line 14: even if it looks worse, reckon it'd be better expected behavior to use 0 as the default padding

GlazeWM.Bar/Components/CPUStatsComponentViewModel.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Reactive.Linq;
+using GlazeWM.Domain.UserConfigs;
+using GlazeWM.Infrastructure.WindowsApi;
+
+namespace GlazeWM.Bar.Components
+{
+  public class CPUStatsComponentViewModel : ComponentViewModel
+  {
+    private CPUStatsComponentConfig _config => _componentConfig as CPUStatsComponentConfig;
+    private readonly CPUStatsService _cpuStatsService = new();
```
lars-berger, line 11: to avoid creating a new instance of the service per component, get it from DI instead:

```csharp
    private readonly CPUStatsService _cpuStatsService = ServiceLocator.GetRequiredService<CPUStatsService>();
```

GlazeWM.Domain/UserConfigs/BarComponentConfigConverter.cs
```diff
@@ -53,6 +53,21 @@ public class BarComponentConfigConverter : JsonConverter<BarComponentConfig>
...
+        JsonSerializer.Deserialize<CPUStatsComponentConfig>(
+          jsonObject.RootElement.ToString(),
+          options
+        ),
+        "gpu stats" =>
+          JsonSerializer.Deserialize<GPUStatsComponentConfig>(
+            jsonObject.RootElement.ToString(),
+            options
+          ),
+        "ram stats" =>
```
lars-berger, line 66: naming the components `cpu`, `gpu`, and `ram` seems more consistent with what we've got currently (eg. we have a `battery` component rather than `battery stats`). this is also in-line with the naming from polybar

Issue #79, opened 2022-06-03
Restore window/workspace state on next startup after a forceful exit


tgharib:
GlazeWM just froze on my PC, which is fine. I don't mind occassionaly restarting GlazeWM. Unfortunately, after having needed to kill it in Task Manager, all my windows except one window are hidden. I suppose this is because before the crash, I was on a workspace with only one window. After starting GlazeWM again, all the hidden windows remain hidden and I essentially lost all my unsaved progress.

There should be a more graceful way of handling crashes. Maybe unhide all windows on GlazeWM start? [nog](https://github.com/TimUntersberger/nog) actually saves all windows in a text file so on restart, not only does it unhide hidden windows but it also puts them in the right workspaces.

tgharib:
If you find yourself in the same situation as I was where unsaved progress was hidden, then [WinLister by NirSoft](https://www.nirsoft.net/utils/winlister.html) has the ability to unhide hidden windows.

lars-berger:
I agree forceful exits should be handled better - there shouldn't be a need to use a 3rd party tool to handle these shortcomings.

Killing the WM via Task Manager unfortunately prevents it from performing any cleanup. I've had the same happen before when a managed window goes fully unresponsive, since the Windows API methods for moving/resizing windows are synchronous and will wait until completion. Is there an error log in `C:\Users\<YOUR_USER>\.glaze-wm\errors.log`?

Many background processes actually make use of hidden windows (it's common for 100+ hidden windows to be running at once), so unhiding all windows will likely cause these to freak out. The idea of saving window/workspace state to a text file sounds pretty solid though. I think it would mean updating the text file after every state-altering command, and there's also the edge case where the user config is changed before the next startup. A cool side benefit of implementing something like this is that it would make it easy to implement a user config reload feature further down the line.

tgharib:
> Is there an error log in `C:\Users\<YOUR_USER>\.glaze-wm\errors.log`?

Unfortunately, losing unsaved progress after a crash is a deal-breaker for me. I use this for work and I can already imagine being in a meeting, doing a screen-sharing presentation and one crash later, all my windows are hidden and I have to fiddle around in WinLister like an idiot. So after I uninstalled, I reflexively removed `~/.glaze-wm`. Opps.

I should add that [komorebi](https://github.com/LGUG2Z/komorebi) tackles this issue by simply minimizing windows rather than hiding them. I'm not implying this method is superior but I figured listing more possible solutions is better.

[...]

PR #273, opened 2023-02-26
fix: Restoring windows after GlazeWM has been forcefully closed

s-v-o:
I was trying to implement a restore after GlazeWM was forcefully closed (#79).

Currently only the managed windows (more precisely the window handle) are saved and when GlazeWM is restarted the ManagedWindowCommand is called again for them.

The layout or workspace is not restored, but the user can access the windows again after the restart.

I'm sure there's a better solution, but maybe this will do until someone comes up with a better one.


lars-berger:
Just a heads up - waiting a bit to review this.

Another member is drafting up a way to do IPC, which opens up another approach of doing this. The idea is to have a separate watchdog process that does 2 things: it keeps track of which windows are managed, and on process kill, it'll ensure all managed windows are shown. One big advantage with doing it that way is the user wouldn't have to relaunch the WM to un-hide the windows. IPC will probably be merged in sometime soon and then it'll possible to test whether that alternate approach is actually gonna be feasible

b1scu1t:
I got a couple bugs with this. 
It saves the managed windows in `~/.glazewm/managedWindows.json' as expected but it would show a warning message stating that 'the file is in use'.
It happens every time GlazeWM is (re) started until the file is manually removed. 
image
With that message, GlazeWM freezes/becomes unresponsive to all Window Events or Keypresses. 
No rules get applied either. It must be forcefully killed in Task Manager.

b1scu1t:
> I hope this will work better for you. But as @lars-berger said, this is more of a workaround and will probably be solved with another solution.


@s-v-o A workaround is better than nothing!
Your merge works as expected, killing GlazeWM unexpectedly isn't as dangerous as before.
The hidden windows return when it is restarted.
I no longer have to painstakingly unhide windows with GuiPropView or WinLister; the experience is leagues better now!

s-v-o:
There is a better workaround now, so I think this pull request could be closed without merging.

Issue #151, opened 2022-10-21
Be able to switch to previously focused workspace


gtsherman:
Just wondering if there's a way to bind a key to switch to the previously focused workspace? I didn't see this in the list of commands but wanted to make sure I'm not missing anything.

Loving GlazeWM so far. Thank you!

azinsharaf:
this is a very necessary feature for any WM. 
 I am referencing the same feature in `Workspacer` WM. https://github.com/workspacer/workspacer/issues/337

s-v-o:
I think this would be a very helpful feature. I find this incredibly useful on i3wm / sway (workspace back_and_forth) especially in the [auto-back-and-forth](https://i3wm.org/docs/userguide.html#workspace_auto_back_and_forth) variant: When pressing the current workspace number it switches back to the last focused one.

kuznetsovvj:
@lars-berger Again, I did not include a new command in the default configuration - I do not know what key combination to use.

kuznetsovvj:
feature is done
@lars-berger maybe close this issue?

lars-berger:
This is now a part of the [latest release](https://github.com/lars-berger/GlazeWM/releases)! 🎉

PR #274, opened 2023-02-26
feat: add option to toggle workspace on refocus

s-v-o:
Added a configuration option to enable automatic back-and-forth (as in [i3wm](https://i3wm.org/docs/userguide.html#workspace_auto_back_and_forth))

With this configuration enabled the previous workspace will be focused when switching to the current workspace.

See also #151

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ general:
   show_floating_on_top: false
   floating_window_move_amount: "5%"
   # When enabled, switching to the current workspace activates the previously focused workspace
-  workspace_auto_back_and_forth: false
+  toggle_workspace_on_refocus: false
 ```
 
 ## Keybindings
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -62,10 +62,20 @@ The configuration file for GlazeWM can be found at `C:\Users\<YOUR_USER>\.glaze-
...
 ```
 
+## General
+
+```yaml
+general:
+  show_floating_on_top: false
+  floating_window_move_amount: "5%"
+  # When enabled, switching to the current workspace activates the previously focused workspace
+  workspace_auto_back_and_forth: false
```
lars-berger, line 76: we don't have to copy i3's naming here. imo we can improve this naming, eg. `back_and_forth_on_workspace_refocus`

the reason it's named that way in i3 is because it'll automatically call a command called `workspace back_and_forth` when the current workspace is refocused, whereas our equivalent command is `focus workspace recent`.

s-v-o: To be honest, I didn't find the i3 naming intuitive either. But I stuck with it because I couldn't find anything better, and so other users coming from i3 will know what it means (and could find some documentation).

If we want to go in a different direction, maybe `toggle_workspace_on_refocus` is even better?

s-v-o:
> Thank you, this is a much needed feature 👍

Thanks to @gtsherman it was really easy to implement.

PR #278, opened 2023-03-01
feat(infra): add helper for getting system media information

AreamanM:


GlazeWM.Bar/GlazeWM.Bar.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <OutputType>Library</OutputType>
-    <TargetFramework>net6.0-windows</TargetFramework>
+    <TargetFramework>net6.0-windows10.0.17763.0</TargetFramework>
```
lars-berger, line 5: would it work to instead specify a [minimum supported version](https://learn.microsoft.com/en-us/dotnet/standard/frameworks#support-older-os-versions)? ie.

```xml
    <SupportedOSPlatformVersion>10.0.17763.0</SupportedOSPlatformVersion>
```

i'm not 100% sure which is better, but it seems like this ^ would give us warnings when we use incompatible APIs

.vscode/tasks.json
```diff
@@ -21,7 +21,7 @@
         "publish",
         "${workspaceFolder}/GlazeWM.Bootstrapper/GlazeWM.Bootstrapper.csproj",
         "--configuration=Release",
-        "--runtime=win-x64",
+        "--runtime=win10-x64",
```
lars-berger, line 24: do we need to change this?

.vscode/tasks.json
```diff
@@ -2,16 +2,16 @@
...
-      "type": "process",
-      "args": [
-        "build",
-        "${workspaceFolder}/GlazeWM.Bootstrapper/GlazeWM.Bootstrapper.csproj",
-        "/property:GenerateFullPaths=true",
-        "/consoleloggerparameters:NoSummary"
-      ],
-      "problemMatcher": "$msCompile"
+        "label": "build",
+        "command": "dotnet",
```
lars-berger, line 6: nit: spacing here should be 2

GlazeWM.Bar/GlazeWM.Bar.csproj
AreamanM: I think it should work, warnings for incompatible APIs sounds great so I'll use it instead of TargetFramework if it builds

AreamanM: Ok so I just tried this out, we still need the explicit windows version in `<TargetFramework>` regardless of the SupportedOSPlatformVersion, and from the docs:

>If you don't explicitly specify this minimum run-time OS version in the project, it defaults to the platform version from the TFM.

so I think it's redundant here

.vscode/tasks.json
AreamanM: just checked; we don't

.vscode/tasks.json
AreamanM: fixed

PR #280, opened 2023-03-05
add execute as user

momozahara:
in case that you have glazewm run as admin and you dont want to execute process and admin

command: exec_a
just a temporary name change to anything you like


momozahara:
oops did I just remove sln?

momozahara:
let me put up new one

PR #281, opened 2023-03-05
add exec_a run process as user and yaml variables support

momozahara:
In case that you have GlazeWM run as admin and you does not want to run process such as powershell as admin
and yaml variable support

```yaml
globals:
- name: "username"
  value: "user_1"
- name: "password"
  value: "12345"
- name: "inner_gap"
  value: 20

gaps:
  inner_gap: "{{inner_gap}}"

command: "exec_a {{username}} {{password}} powershell"
```


lars-berger:
IMO the WM shouldn't be run as admin. It's necessary right now if you want to manage elevated windows, but it should actually be possible to manage elevated windows as a non-elevated process if we do [this](https://github.com/lars-berger/GlazeWM/issues/32#issuecomment-1100632889). It'll likely be at least a month+ before we get a code signing certificate, but that's the plan for now.

Issue #276, opened 2023-02-28
reserved space?


calendulish:
I used to reserve some space on right side of my desktop to widgets, is there a way to tell glazewm to ignore that area when tiling the windows?

something like right_gap or right_margin

lars-berger:
We could tweak `gaps.outer_gaps` to optionally accept multiple values like we do for properties like `bar.padding`,  eg.

```yaml
gaps:
  outer_gap: "20px 40px 20px 20px"
```

If you wanna take a stab at a PR, the way to go would be to change `OuterGap` in `GapsConfig.cs` to a string and then adjust the height/width/x/y of `Workspace`. `UnitsHelper.cs` can help with converting the string to ints

PR #282, opened 2023-03-06
feat: add shorthand syntax to outer gaps

maxle5:
* Added support for specifying individual outer gaps (top, right, bottom, left)
* Also tested that this change is backwards compatible with existing config (i.e. `outer_gap: 20` still works)

closes: #276


lars-berger:
Thank you for this 🙏 And sorry about taking absolutely ages to get this merged in - will be putting together a new release with this included in the next couple weeks

PR #284, opened 2023-03-13
feat: improve handle management

programmer2514:
Refactors the `IsHandleManageable()` routine.

This allows for implicit compatibility with Flow Launcher and improves code conciseness by replacing the unreliable "menu window" check with the more robust `WindowStyles.Popup` flag.

Additional changes:
  - Removed unnecessary dependency from `FocusContainerUnderCursorHandler.cs`
  - Updated the solution to be compatible with Visual Studio 2022 (shouldn't break compatibility with previous versions)

### DIFF
--- a/GlazeWM.Domain/Containers/CommandHandlers/FocusContainerUnderCursorHandler.cs
+++ b/GlazeWM.Domain/Containers/CommandHandlers/FocusContainerUnderCursorHandler.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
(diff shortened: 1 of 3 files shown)


calendulish:
It causes problems with some particular applications (like Steam, Whatsapp Beta) making them unmanageable

lars-berger:
Due to the C# -> Rust rewrite, some of these older PRs aren't compatible anymore. Can't merge as is, and I'm sorry about not finding the time to get this in while the C# version was active. Nevertheless, thank you for contributing 🙏

PR #303, opened 2023-04-20
Fix workspace toggle

s-v-o:
@Petar0000 mentions [two issues](https://github.com/lars-berger/GlazeWM/pull/274#issuecomment-1507028934) with PR #274 that this PR fixes.


lars-berger:
Thank you for fixing this 👌 

Will be putting together a release with this included in the next couple weeks

PR #338, opened 2023-06-30
feat: alias Ctrl to Control

JK-Flip-Flop96:
Tiny PR here. 

I grabbed some low-hanging fruit from the "Free to take" list. https://github.com/users/lars-berger/projects/2/views/1?pane=issue&itemId=28116872

I chose to implement this with a switch so that more keybinding aliasing could be done quickly in future (e.g. Meta -> Alt)

Ignore the completely overwritten first commit, I decided as I was drafting this PR that it'd be better to check against the Enum first before moving on to the switch so there's not a whole bunch of unnecessary comparisons.


lars-berger:
Thank you for this 🙏

PR #351, opened 2023-08-08
fix: compilation error in `VolumeComponentViewModel`

maxle5:
I was unable to compile the solution after some PRs were merged over the last few days. Looks like one of the methods is now expecting a `Dictionary<string, Func<string>>` instead of `Dictionary<string, string>`


lars-berger:
Good catch, thank you 👍

PR #382, opened 2023-09-06
feat: inspect window command

maxle5:
## Description
Allows for viewing information about the window under the user's cursor (e.g. process name, class name, title, etc.)

## Motivation/Context
I couldn't find an easy way to identify the "class name" of a window (without third-party tools), so I decided to implement a very basic command that shows a dialog with this information. This could be expanded upon in the future to show more info such as height, width, etc.

## Demo
https://github.com/glazerdesktop/GlazeWM/assets/8741942/8bedb169-5f3e-450d-87d1-32c76046cd88


lars-berger:
Ay this is really cool, thank you 🙏. There was some discussion not too long ago about an alternative approach to solving more or less the same issue here https://github.com/lars-berger/GlazeWM/issues/361 But the way you've solved this is more user friendly compared to a CLI. 

I think it'd be worth making a slight tweak here to cover some more use cases where a user would need to know a window's class name etc. One case is for windows that aren't being managed by the WM (eg. if we add a "force manage" command in the future), and another is for popups that are shown briefly. We could cover these cases by instead inspecting the window currently under the cursor.

If you're cool with making this change, `FocusContainerUnderCursorHandler` has some snippets that would be useful

maxle5:
@lars-berger, yeah that totally makes sense! Hadn't really thought of those other use cases, I'll give that a shot sometime this week and see what I can come up with

lars-berger:
@maxle5 Sweet 👌  Feel free to message in #contributing for any questions etc

maxle5:
@lars-berger, sorry it took me so long to get back to this, but this PR should now cover the scenarios you described above by displaying the window information of the window under the user's cursor.

Let me know what you think, or if there's anything you think I should change.

lars-berger:

PR #422, opened 2023-10-04
feat: automatic tiling direction

phisko:
Option to automatically set the tiling direction when a new window is focused or resized
* unchanged
* horizontal
* vertical
* alternate between horizontal and vertical
* automatically pick one according to the window's largest dimension (height vs width)


lars-berger:

PR #430, opened 2023-10-16
fix: correct workspace size with Y offset and bar on bottom

silvarc141:
Fix workspace size and Y position when using `offsetY` bar config option with bar on the bottom


lars-berger:
Sweet, thanks for fixing this 😄 Looks good

PR #459, opened 2023-12-09
docs: add readme section for using `LWin` keybindings

Sebastian-Nielsen:
I added a guide on how to remap `LWin + <somekey>` to use with GlazeWM.

Feel free to move this out of the README.md if it takes up too much space


lars-berger:
Sweet, thank you 🙌 At some point the docs will need to get a touch up and get separated into pages. But this gets asked all the time, so it'd definitely be useful to have some docs on how to achieve it. Genuinely didn't know you could preserve lock functionality while still remapping `Win+L`

PR #467, opened 2023-12-13
feat: add focus monitor next/prev command

freestingo:
added new command for moving the focus sequentially between different monitors.


lars-berger:
Hey, thanks for the PR 🙌The actual code looks fantastic but I'd suggest we change from `focus monitor <prev | next>` to `focus monitor <left | right | up | down>`.

The monitor sequence is more of an internal implementation and they aren't guaranteed to be in a left-to-right order. And I can imagine it could be confusing to determine next/prev for people who have unorthodox setups with monitors stacked vertically etc.

We currently don't have a command for focusing to a monitor in direction, and it'd help with getting more i3wm features moved over. If you're down to make the change, lmk 👍 We have a utility for getting a monitor in a given direction via `MonitorService.GetMonitorInDirection`.

freestingo:
@lars-berger just pushed a new version that supports focusing monitors in a given direction according to your feedback :D

lars-berger:

PR #471, opened 2023-12-14
fix: make `FocusInDirectionCommand` a no-op if focused container is maximized

### LINKED ISSUE #469: `focus <direction>` commands e.g `focus left` should no-op when current focused window is maximized
I use Glazewm compiled from the `develop` branch as as of commit #421 
Currently, if e.g. `focus left` is executed when you have the current focused window maximized, the window on the left becomes focused and the previously focused window stays maximized and goes behind the newly focused window, but since it is still maximized, it covers all the rest of the windows on the workspace.

wis:
fixes #469


lars-berger:
Sweet, thanks for fixing this 🙌

PR #472, opened 2023-12-14
fix(cursor_follows_focus): do not move cursor to 0,0 when focusing workspace with a maximized window

### LINKED ISSUE #470: [Bug] when `cursor_follows_focus: true` and you move to a workspace with a maximized window, cursor moves to (0, 0) not center
I use Glazewm compiled from the develop branch as as of commit #421

wis:
fixes #470


lars-berger:

PR #475, opened 2023-12-14
fix: system tray component expand on mouse button up

Video-Nomad:
I've noticed that some system tray icons are getting triggered on expand if that icon is right under the expand arrow.
Changing the event trigger to MouseLeftButtonUp fixed the issue.


lars-berger:
Dealing with the quirks of WPF is a pain, so thank you for taking care of this 👌

Issue #436, opened 2023-10-22
[Feature request] A floating command and rule that does not tile the window first, like set floating is doing now.


Video-Nomad:
Hi,

It would be great to be able to create a window rule that will launch the new app in floating mode by default, without first docking it and undocking.

An example of the issue I currently have with "set floating":
- I have two apps tiled side by side, widths are 30/70.
- I open a new app that has "set floating" in window_rules.
- The expected behavior is that the 30/70 width is maintained in the original tiles and a new window is simply created on top as floating.
- The actual behavior is that the new app is first docked, and then undocked, messing with the widths of the original tiles in the process.

Setting the tiling direction to vertical has no effect on this behavior.

As an additional note, "set floating" always centers the window[...]
<details>
<summary>[FIXED] with #476 on develop branch. Floating centering is now optional. </summary>

[...] but it would be better for some apps to have an option to use the original position. For example, when I use Autodesk Maya ("set floating" by default) all the in-app floating windows are always centered by GlazeWM, and you open and close those windows all the time when using the app. Setting those apps to "ignore" is not great either, since you can't set to ignore and tile some of those windows if you need to later, there's no ignore toggle.

Using regex title match works in this case, so I can set to ignore everything except for the main window that has "Autodesk MAYA":
```
- command: "ignore"
    match_process_name: "/maya/"
    match_title: "/^(?:(?!Autodesk MAYA).)*$/"
```
But it might not work with other similar apps.
</details>

Really enjoying the window manager, the best I've tried, that's for sure. Thank you!

nahid6970:
same issue  have u found any solution

lars-berger:
Fixed in v3 👍 [Latest release](https://github.com/glzr-io/glazewm/releases)

PR #476, opened 2023-12-14
feat: option to disable floating windows initial centering

Video-Nomad:
Issue: some apps (Autodesk Maya, 3ds Max) have a workflow of creating floating windows constantly. Setting a rule of "set floating" to those windows will center them on the screen every time you open them. By default, those apps keep the position of those windows internally, but GlazeWM will treat them as new windows every time and center them (even within one session).

Solution: have an option to disable initial centering of newly created floating windows. Default behavior will not change, unless the user decides to set the option to false. I keep it constantly to false in my build.

Possible issues: I haven't tested this on a multi-monitor setup, as I don't use multiple screens.

Partially closes the (#436)

### DIFF
--- a/GlazeWM.App/Resources/sample-config.yaml
+++ b/GlazeWM.App/Resources/sample-config.yaml
@@ -15,13 +15,13 @@ general:
   # Amount to move floating windows by (eg. when using `alt+<hjkl>` on a floating window)
   floating_window_move_amount: "5%"
 
+  # Whether to center new floating windows.
+  center_new_floating_windows: true
+
   # *Strongly* recommended to set to 'false'. Whether to globally enable/disable
   # window transition animations (on minimize, close, etc). Set to 'unchanged'
   # to make no setting changes.
   window_animations: "unchanged"
-
-  # Whether to center new floating windows (default is true)
-  center_new_floating_windows: true
 
 gaps:
   # Gap between adjacent windows.
--- a/GlazeWM.App/Resources/sample-config.yaml
+++ b/GlazeWM.App/Resources/sample-config.yaml
@@ -20,6 +20,9 @@ general:
   # to make no setting changes.
   window_animations: "unchanged"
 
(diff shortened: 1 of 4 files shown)


lars-berger:
Hey, thanks for the PR! 👍 Started a poIl in the discord just now for whether we should change the default to `false` instead. I actually think more people find the centering annoying than useful, but we can wait a few hours before making the change

There's a minor issue when using multi monitors. If you have an empty workspace 1 on monitor A and a non-empty workspace 2 on monitor B, then when a window is opened with workspace 1 focused, the OS will plop it on workspace 2. So there just needs to be a check in `ManageWindowHandler.cs` whether the window was opened on a different monitor than where it should go, and if so, translate it to center.

Video-Nomad:
Thanks for the heads-up @lars-berger ! I need to buy a second monitor, I guess 😅 
I think this check should do the trick. If the window is not in the correct workspace it will snap it to the center:
```csharp
var targetWorkspace = WorkspaceService.GetWorkspaceFromChildContainer(targetParent);
var windowWorkspace = _monitorService
  .GetMonitorFromHandleLocation(windowHandle)
  .DisplayedWorkspace;

var floatingPlacement = _userConfigService.GeneralConfig.CenterNewFloatingWindows
  ? originalPlacement.TranslateToCenter(targetWorkspace.ToRect())
  : originalPlacement;

if (targetWorkspace != windowWorkspace)
  floatingPlacement = floatingPlacement.TranslateToCenter(targetWorkspace.ToRect());
```
Let me know if it will work, I can't really test it on my end at the moment.

Also, how's the poll going? Should I change the floating centering to be false by default?

lars-berger:
@Video-Nomad Surprisingly it looks like more people prefer the default centering behavior (15:4 votes so far), so it's fine to leave as is 👍 

Thanks for the snippet, I can test it out later and commit it directly to your branch

Issue #78, opened 2022-05-26
monocle layout support?


ut4k:
Will this software support monocle layout in the future?
If so, it could be the best wm for windows for me.

lars-berger:
Is monocle layouts a feature from bspwm? 

Looking to add some sort of fullscreen toggle feature in the future, ref [here](https://github.com/lars-berger/GlazeWM/issues/70#issuecomment-1124342503). The implementation is likely to be similar to the `fullscreen enable` command in i3.

tgharib:
> The implementation is likely to be similar to the `fullscreen enable` command in i3.

AFAIK, equivalent of monocle layout in i3 would be `layout tabbed` instead of a simple fullscreen.

ericreeves:
I have no idea what this is called in i3, but ultimately the behavior is:

When monocle mode is enabled, it behaves exactly as if only a single application is open in a workspace, so the application takes up the whole screen (excluding configured gaps), but then you can still use next/prev app to cycle through all apps open on that workspace.  Disabling monocle mode restores windows to their previous locations.

It doesn't use proper "full-screen" or "maximize", so it should be far simpler to implement than #70 .

Thanks for your work on GlazeWM!  Given all of the TWM-like options on Windows, I feel it's the best one so far!

b1scu1t:
I think a nice way to implement this only checking 'managed' (not floating/ignored) windows and minimizing/hiding anything but the focused window. Though...this way GlazeWM would have to remember the window states *before* Monocle is activated.

ut4k:
what i want be like:
[https://i.imgur.com/oVatUvy.png](https://i.imgur.com/oVatUvy.png)

freestingo:
+1 for this layout; it's the one I use pretty much all the time in `xmonad` and would really love to see it also implemented here

lars-berger:
Closing in favor of #678

PR #478, opened 2023-12-16
Feat: monocle mode

Fuann-Kinoko:
Hi, I've tried to add a simple monocle implementation #78 that doesn't involve any modifications to the existing tiling window logic.

Instead of defining new monocle containers, it just draws each tiling window (floating windows are ignored) in the workspace where `isMonocle` is set to true, to fit most of the screen (the same size as the rectangle of a tiled window with a 100% size percentage). If the `isMonocle` flag is set to false, then GlazeWM restores the position of the managed windows correctly, since they were never touched or modified in monocle mode.

And because there are only two directions in Monocle mode: next/prev, I've also added two commands: `focus next/prev` to focus the corresponding window in the nested window tree (they're essentially the same as focusInDirection, but less restrictive), and an override of hjkl to focus next/prev when the workspace is in monocle mode.

This implementation can be tested by adding a keyboard binding to the command `toggle monocle` (`focus next` and `focus prev` is also added).


Fuann-Kinoko:
bar component `monocle indicator` added for convenience

```
    - type: "monocle indicator"
      label_entered: "●"
      label_exited: "○"
      foreground: "#b4637a"
```

lars-berger:

PR #484, opened 2023-12-22
fix: workspace shouldn't be flattened

Fuann-Kinoko:
Imagine a workspace with three windows.

fig1

Closing (detaching) TW1 triggers the following section: 

https://github.com/glazerdesktop/GlazeWM/blob/18eb8c80ceaf166eaf6c74b7a3846871e18db9ae/GlazeWM.Domain/Containers/CommandHandlers/DetachContainerHandler.cs#L70-L80

in this case, `detachedSiblings[0]` refers to the `splitContainer`, and `detachedParent` is the workspace. The `splitContainer` flattens correctly, but then flattening the workspace causes its deletion, directly attaching `TW2` and `TW3` to the monitor.

Then during redrawing, `TW2` and `TW3` can't find their workspace ancestor, leading to the exception `Sequence contains no elements`.
https://github.com/glazerdesktop/GlazeWM/blob/18eb8c80ceaf166eaf6c74b7a3846871e18db9ae/GlazeWM.Domain/Containers/CommandHandlers/RedrawContainersHandler.cs#L79

The fix is just adding a conditional check to whether `detachedParent` is a workspace. But there might be potential problems that might arise from executing the flatten command.

### DIFF
--- a/GlazeWM.Domain/Containers/CommandHandlers/DetachContainerHandler.cs
+++ b/GlazeWM.Domain/Containers/CommandHandlers/DetachContainerHandler.cs
@@ -75,8 +75,8 @@ namespace GlazeWM.Domain.Containers.CommandHandlers
         _bus.Invoke(
           new FlattenSplitContainerCommand(detachedSiblings.ElementAt(0) as SplitContainer)
         );
-
-        _bus.Invoke(new FlattenSplitContainerCommand(detachedParent as SplitContainer));
+        if (detachedParent is not Workspace)
+          _bus.Invoke(new FlattenSplitContainerCommand(detachedParent as SplitContainer));
       }
 
       return CommandResponse.Ok;


lars-berger:
Was just about to put out a small new release, so this came in right in time 👍 Looks good to me

lars-berger:
Btw I haven’t forgotten about your monocle PR. There’s a general idea for how we can add support for layouts as a whole in GlazeWM in the future, and I’m trying to figure out a way where we can add monocle mode in rn and not have too many breaking changes when we add other layouts. I’ll get back to you with that one in a day or two

Fuann-Kinoko:
👌Of course, I'm perfectly fine if you choose to leave that PR as is or close it.

Personally I find that the implementation is a bit forceful, as it was designed to address an immediate need and might not be the best fit for inclusion in the master branch. 

If GlazeWM introduces layouts in the future, perhaps monocle mode would likely be the simplest one to integrate as well.

Issue #456, opened 2023-12-07
[Bug] Parsec affects other windows / tiling behavior


stianeklund:
I'm not sure if this strictly speaking is a bug or rather an issue with Parsec.

1. I'm not able to move the Parsec window around (I'm guessing this might be due to input capture behavior?).
2. Opening Parsec or returning from a minimized state on one workspace affects the tiling of neighboring workspace.
3. If I change focus from say, the windows file explorer to Parsec (already being open at this point) I can't move focus away from Parsec / can't alt+l / alt+h to change focus to another application once Parsec has gained input focus.

I've attached a screen recording demonstrating the interaction behavior with Parsec. Please ignore the black box in the video as it's something I added to remove my username from the video.

https://github.com/glazerdesktop/GlazeWM/assets/4148195/bad1d4ad-b82d-4fb2-be14-bfddffd12ce0

P.s. I checked the error log file in `.glaze-wm` and can't find any errors there.

I can potentially debug it to help further but I'm curious whether or not this is simply a "feature" or actually a bug.
I suspect if Parsec intercepts key input then it might not be possible to fix.

stianeklund:
I did some further testing and believe elevation isn't the issue at least not when it comes to interacting with Parsec. Parsec isn't the only app that behaves this way, certain Windows apps like Event Viewer. Specifically, Event Viewer and similar apps have conflicting key bindings which are some of the issues but IMO the larger problem is that GlazeVM doesn't handle/tile the window at all.

One workaround I've fun at least for Parsec is to run Parsec through [Sandboxie](https://sandboxie-plus.com). I have no idea how or why this works. I have no insight into GlazeVM's code base & how other software handles interaction like this through the Windows API.

Issue #331, opened 2023-06-18
system tray isn't showing all icons


calendulish:
I noticed that when activating the system tray in GlazeWM, some tray icons do not appear. This also prevent the icons to appear on the windows taskbar, so I need to disable the component and restart the programs to these icons appear.

aaronedev:
> From my tests, when running GlazeWM as admin, it only shows tray icons of programs also running as admin. When Glaze is not running as admin, all icons are visible (also those running as admin).

yea that is correct - for me it behaves the same but whenever i open the terminal it wont focus it kinda floats and fucks up the alignment of the windows. Anyone experiences the same and does know how to fix it?

zvbt:
> > From my tests, when running GlazeWM as admin, it only shows tray icons of programs also running as admin. When Glaze is not running as admin, all icons are visible (also those running as admin).
> 
> yea that is correct - for me it behaves the same but whenever i open the terminal it wont focus it kinda floats and fucks up the alignment of the windows. Anyone experiences the same and does know how to fix it?

i had the same issue its when you open an app in admin for my terminal i used this in my config

```
- command: "set floating"
    match_process_name: "alacritty"
```

lars-berger:
I'm closing feature requests and bugs about the built-in bar since it's now been swapped out for [Zebar](https://github.com/glzr-io/zebar). If Zebar doesn't fix what you need, a new issue for it can be opened over [here](https://github.com/glzr-io/zebar/issues)

PR #485, opened 2023-12-22
feat: make elevating process in execution an option

Fuann-Kinoko:
If `GlazeWM` isn’t run with administrator privileges, it can't handle certain windows properly (#456 ). However, running `cmd.exe` or any other programs in an elevated `GlazeWM` will cause it to be opened with elevated permission as well.

I've attempted a (somewhat ugly) trick by using `explorer.exe` to open a shortcut of the program, allowing the target process to remain unelevated.

But there are still two aspects I believe need consideration:
1. Is it necessary to delete the created `temp.lnk`? Although launching the new program occurs in a separate process, to avoid potential issues, `GlazeWM`'s main process still waits for `explorer.exe` to finish before deleting `temp.lnk`. This might slightly increase program startup time. If `temp.lnk` isn’t deleted (I'm unsure if users would be concerned about it), none of the aforementioned issues would arise.
2. What’s the default option for running programs? Currently, my approach is that `exec cmd.exe` spawns an unelevated process, while `sudo exec cmd.exe` invariably spawns an administrator process, regardless of whether `GlazeWM.exe` has elevated privileges.

### DIFF
--- a/GlazeWM.Domain/Common/CommandHandlers/ExecProcessHandler.cs
+++ b/GlazeWM.Domain/Common/CommandHandlers/ExecProcessHandler.cs
@@ -1,6 +1,6 @@
 using System;
-using System.IO;
 using System.Diagnostics;
+using System.IO;
 using System.Windows.Forms;
 using GlazeWM.Domain.Common.Commands;
 using GlazeWM.Infrastructure.Bussing;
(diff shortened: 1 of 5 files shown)


Video-Nomad:
Runing GlazeWM as admin also causes the system tray component to ignore all non-elevated apps #331

stianeklund:
@Fuann-Kinoko I don't think running GlazeVM as admin is the best option from a security standpoint, however, is there any way to tell if the window is an elevated one? If so maybe UAC elevation could be nice to allow interaction? I don't have any insight into this so forgive me if I'm asking in the wrong place & wrong person etc.

lars-berger:

PR #528, opened 2024-02-01
Added 'set no_title' command to allow the user to remove the title bar from windows.

FrancisTheCat:
This includes minor changes to WindowsApiService.cs, because when using it, I found the SetWindowLong function unaccessible on 64 bit machines, which is wrong, since it actually does not consider 32 bit pointers, but just values wich are 32 bit even on 64bit systems. This might break compatibilty with 32 bit systems, but Id be surprised to hear if that compatibilty existed in the first place.

Note: This unfortunately makes the window no longer resizeable with the mouse due to Windows' **great** design, but I dont think thats a big issue since this is a tiling window manager lol.


lars-berger:

FrancisTheCat:
> Due to the C# -> Rust rewrite, some of these older PRs aren't compatible anymore. Can't merge as is, and I'm sorry about not finding the time to get this in while the C# version was active. Nevertheless, thank you for contributing to GlazeWM 🙏

All good, I totally understand that. I probably wont make an equivalent pr on the rust version, tho since I aint waiting for that to compile lol, tho I hope this pr may still be of use to someone if they want to implement it in the new version, since I do think its quite a nice feature.

PR #531, opened 2024-02-04
feat: add click commands to text file component

DaNubCoding:
Please excuse me and point out if there are issues with the way I wrote the code, as I have never programmed in C# before.


lars-berger:

PR #534, opened 2024-02-06
Add json schema for config

EmilyGraceSeville7cf:


lars-berger:
@EmilyGraceSeville7cf Wow, thanks for taking the time to do this 🙌  I've used JSON schemas for JSON files (eg. tsconfig, prettier etc), but have never used them for YAML before. How do I use these as an end user?

Also, would you happen to know whether SchemaStore accepts YAML configs like this one?

EmilyGraceSeville7cf:
I finished my PR. I need a review, because I have some doubts about what properties should be marked as required (it's not documented).

EmilyGraceSeville7cf:
@lars-berger, can you review this PR?

JustBarnt:
I noticed the schema has an error with `workspaces/bind_to_monitor` its marked as a boolean when it should be an integer?

EmilyGraceSeville7cf:
@JustBarnt, I've fixed this problem. :smi

HeyItsGilbert:
Love this! FYI Lars there aren't YAML schemas, they just use JSON schemas. I was totally caught off guard when I first learned that. YAML considers JSON as valid subset of YAML.

mathwro:
I just did some testing with schema and found it to somewhat out of date and updated in YAML for anyone interested :)

https://gist.github.com/mathwro/dfe1c62d885cd39fb98adb72f798324c

PR #552, opened 2024-03-03
Add correct volume mute icon in README.md

JohannesPertl:


lars-berger:
Due to the C# -> Rust rewrite, some of these older PRs aren't compatible anymore. Can't merge as is, and I'm sorry about not finding the time to get this in while the C# version was active. Nevertheless, thank you for contributing🙏

Issue #554, opened 2024-03-05
Cursor Follows Focus - Cursor does not Follow when Workspace has no Windows


blampewpew:
I am using GlazeWM in conjunction with Flow Launcher. I have `cursor_follows_focus: true` set in my config and have a multiple monitor set up with workspaces set to each monitor. 

The cursor following focus works as expected between workspaces that have active windows open. However, when a workspace has no windows opened, the cursor does not follow.

Am wondering if it is possible to implement or if I'm missing any config settings to allow this behavior since I believe it will help my issue with launching the Search Window in an empty workspace with Flow Launcher.  

Please let me know if I'm missing anything required for creating an issue.

lars-berger:
This has been fixed now in v3 👍 [Latest release](https://github.com/glzr-io/glazewm/releases)

PR #557, opened 2024-03-15
feat: Allow Mouse to Follow Cursor onto Empty Workspace and Simulate Mouse Click to Facilitate Launching Flow Launcher Search

blamcodes:
Here's my attempt to address Issue #554 which allows Flow Launcher to launch search on an empty workspace as well as other programs that do not like that funky state of a Workspace that has been focused but not clicked (e.g., Windows Screenshot Tool). 

I'm an avid user of i3 and am very happy that this project exists which allows me to have nearly the same experience on my Windows machine. Thanks and kudos to the devs who started this! 

## Overview of Changes:
- **Enable mouse cursor to follow onto a workspace that is empty**.
  - After debugging the console output, I noticed that when the focus is changed to a workspace that has no active windows, the `FocusedContainer` is a `GlazeWM.Domains.Workspaces.Workspace` type container. I modified the `.Where()` on WmStartup.cs:123 to also invoke the `CenterCursorOnContainerCommand` when this happens. This allows for the cursor to follow onto an empty workspace.
- **Simulates Mouse Click only when Focus is changed to an Empty Workspace**.
  - Even after getting the mouse cursor to follow onto an empty workspace, I was still not able to launch the Search Window of Flow Launcher after focusing. It appears that the empty workspace is not actually focused (?) since I can still see the title of the last focused window in the Polybar even though the Workspace Number in the polybar is showing active.
image
image
(I can't even initiate the Windows Screenshot tool (WIN+SHIFT+S) to get a full screenshot of the highlighted workspace number and the title of the last window on my polybar when in this state so it's not just Flow Launcher that's having issues.)
When I click on the empty workspace, that seems to effectively focus the workspace and the polybar title is cleared and I can launch Flow Launcher (and the Windows Screenshot tool).
  - I added `SendInput` along with the required structs for Mouse Input and added that to the `CenterCursorOnContainerHandler` with a check that the container is of type `Workspace`.

### DIFF
--- a/GlazeWM.App.WindowManager/WmStartup.cs
+++ b/GlazeWM.App.WindowManager/WmStartup.cs
@@ -6,6 +6,7 @@ using GlazeWM.Domain.UserConfigs;
 using GlazeWM.Domain.UserConfigs.Commands;
 using GlazeWM.Domain.Windows;
 using GlazeWM.Domain.Windows.Commands;
+using GlazeWM.Domain.Workspaces;
 using GlazeWM.Infrastructure.Bussing;
 using GlazeWM.Infrastructure.Common;
 using GlazeWM.Infrastructure.Common.Commands;
@@ -119,7 +120,7 @@ namespace GlazeWM.App.WindowManager
             .Select((@event) => @event.FocusedContainer);
 
           focusedContainerMoved.Merge(nativeFocusSynced)
-            .Where(container => container is Window)
+            .Where(container => container is Window or Workspace)
             .Subscribe((window) => _bus.InvokeAsync(new CenterCursorOnContainerCommand(window)));
         }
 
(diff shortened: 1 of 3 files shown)


blampewpew:
Investigating exception being thrown when coming out of the lock screen. Looks like `result` from SendInput is triggered and is `0`.

blampewpew:
I was running an old build where I didn't add the 
```
if (container is Workspace) {}
```
check in the `CenterCursorOnContainerHandler` which was causing the Exception to be thrown.

It's good.

lars-berger:
Due to the C# -> Rust rewrite, some of these older PRs aren't compatible anymore. Can't merge as is, and I'm sorry about not finding the time to get this in while the C# version was active. Nevertheless, thank you for contributing🙏

PR #602, opened 2024-05-08
feat: added DefaultLabel to BindingMode Component

ParasiteDelta:
The motivation is pretty simple: if you use separators for your bar, there will be a blank space between two separators where the binding mode display is. The reason is that while it's set up to collapse the width when there is no binding mode by default, the bar still reserves this space. So, I've added in a config variable, `default_label`, to optionally display a tag whenever the user does not have a binding mode selected.

Let me know if I need to make any changes.


ParasiteDelta:
Closed due to source nuking.

PR #630, opened 2024-07-03
fix: typo in sample config

tjk:


lars-berger:
This has been fixed in the sample config now 👍

Issue #564, opened 2024-03-28
Move windows with mouse


ISAKEM:
Is it possible to move windows between monitors with the mouse like u can do on FancyWM and Komorebi?

txj-xyz:
Not that I have found, I typically move them via shortcut at the moment.

LucaCoduriV:
Looking for the same functionality :(

LucaCoduriV:
@HolbyFPV  @lars-berger 

Hi,

I'm currently exploring how to implement this feature. Although I have much to learn about the Windows API, I’m eager to contribute, especially because I would love to use this feature myself.

Here's a brief outline of how I envision implementing it:

1. Detect when a window is being dragged.
2. Capture the mouse position when the window drag state ends.
3. Calculate the window's new position within the current workspace tree based on the mouse position.
4. Move the window to the new position (handling the interaction with other windows is something I'm still figuring out).

As @HolbyFPV mentioned on Discord, some windows might not have a title bar. In such cases, we can combine a keyboard key with a mouse drag to move the window. Ultimately, the same approach should work for both scenarios.

I’ve looked at your platform code, and it seems we need to handle the WM_ENTERSIZEMOVE event to listen for both window drag and resize exits. (The resize part could be a separate issue, in my opinion).

I haven't delved much further into this yet because I want to ensure no one else is already working on it and to confirm that it’s okay for me to proceed.

Looking forward to your feedback!

lars-berger:
@LucaCoduriV Yeah absolutely, feel free to take a stab at it. The latest development branch is `quick-test` until the Rust rewrite gets merged.

There's a really cool POC that @Pamesta actually did a while back which might be a useful reference ([demo video](https://discord.com/channels/1041662798196908052/1041679140719448124/1080558513111175328)). Their branch is [here](https://github.com/glzr-io/glazewm/compare/main...Pamesta:GlazeWM:mouseMoveAlpha).

There's different ways to approach the window move behavior and I'm not sure what's best. A couple things to consider:
* Should it be possible to go from tiling -> floating when moving a window via drag? If so, when should this trigger?
* Should the insert position be based off where the window is dragged to? e.g. insert at left/right depending on window position
* Should it ever create a new split horizontally/vertically? e.g. create a vertical split if window is near the bottom of another window

IMO ideally we could have an implementation where it's just purely mouse click and drag, but if it really calls for it, it might be useful to have different behaviors if a specific key is held down.

[...]

PR #635, opened 2024-07-26
feat: move window by dragging them with the mouse

LucaCoduriV:
# Goal 
Add the possibility to organize windows using the mouse. #564

# Result:

https://github.com/user-attachments/assets/ab552c1e-4ed1-47cf-8b14-bae828b69519


lars-berger:
Ay awesome work so far 👍 Hmu on Discord if you get stuck on something

packages/wm/src/containers/commands/get_containers_at_pos.rs
```diff
@@ -0,0 +1,32 @@
...
+  common::Point,
+  containers::{
+    traits::{CommonGetters, PositionGetters},
+    Container, WindowContainer,
+  },
+  wm_state::WmState,
+};
+
+/// Return all window under the mouse position 
+pub fn get_containers_at_position(
```
lars-berger, line 11: Commands are for making changes to state, so it'd be better if this was a getter fn on `WmState`

packages/wm/src/common/events/handle_window_location_changed.rs
```diff
@@ -131,3 +145,77 @@ pub fn handle_window_location_changed(
...
+fn set_into_floating(
+  moved_window: TilingWindow,
+  state: &mut WmState,
+  config: &UserConfig,
+) -> anyhow::Result<()> {
+  let moved_window_parent = moved_window
+    .parent()
+    .context("Tiling window has no parent")?;
+
+  if let Some(Container::Split(split)) = moved_window.parent() {
```
lars-berger, line 192: can this be removed? split containers are automatically flattened if empty on window detach

packages/wm/src/common/events/handle_window_moved_or_resized_end.rs
```diff
@@ -0,0 +1,60 @@
...
+
+      if has_window_moved {
+        window_moved_end(window, state, config)?;
+      }
+    } else if let WindowContainer::TilingWindow(window) = window {
+      window_resized_end(window, state, width_delta, height_delta)?;
+    }
+  }
+
+  Ok(())
```
lars-berger, line 59: can we add the add the functions from `window_moved_end` and `window_resized_end` into here?

packages/wm/src/common/events/handle_window_moved_end.rs
```diff
@@ -0,0 +1,323 @@
...
+  dropped_position: DropPosition,
+  parent: &Container,
+  split_container_index: usize,
+) -> anyhow::Result<()> {
+  let target_index_inside_split_container = match dropped_position {
+    DropPosition::Start => 0,
+    DropPosition::End => 1,
+  };
+
+  let split_container = Container::Split(SplitContainer::new(
```
lars-berger, line 240: you can use the existing `wrap_in_split_container` fn here

packages/wm/src/common/events/handle_window_moved_end.rs
```diff
@@ -0,0 +1,323 @@
...
+  );
+
+  let tiling_direction = get_split_direction(&window_under_cursor)?;
+  let new_window_position = get_drop_position(
+    &mouse_position,
+    &window_under_cursor,
+    &tiling_direction,
+  )?;
+
+  let parent = window_under_cursor.parent().unwrap();
```
lars-berger, line 77: can be simplified with `direction_container()` getter which gets the nearest ancestor with `TilingDirection`
```rust
let parent = window_under_cursor.direction_container().context("...");
let tiling_direction = parent.tiling_direction();
```

packages/wm/src/windows/window_operation.rs
```diff
@@ -0,0 +1,8 @@
+#[derive(Debug, Copy, Clone, Default, PartialEq)]
+pub enum WindowOperation{
```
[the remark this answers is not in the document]
LucaCoduriV: Yes but I need the waiting state. I use this state to know that I have to check if it's either a resize or a move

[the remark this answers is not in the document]
LucaCoduriV: Oh sorry didn't saw you comment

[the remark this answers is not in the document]
lars-berger: the `operation` could be `None` while it's not known whether it's a move/resize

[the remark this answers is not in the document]
LucaCoduriV: No because it's none when not dragged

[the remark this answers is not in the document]
LucaCoduriV: I put it on Wainting only when a drag or resize started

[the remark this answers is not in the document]
lars-berger: i'm thinking the `ActiveDrag` struct would be wrapped in `Option` and only exists if the window is being dragged

```rust
pub trait WindowGetters {
  fn active_drag(&self) -> Option<ActiveDrag>;
  ...
```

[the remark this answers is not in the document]
LucaCoduriV: Done

packages/wm/src/containers/commands/get_containers_at_pos.rs
LucaCoduriV: done

packages/wm/src/common/events/handle_window_location_changed.rs
LucaCoduriV: If I want to copy the behaviour of hyprland I think that I have no other choice.
They remove the split container when the window is alone in it.

packages/wm/src/common/events/handle_window_moved_or_resized_end.rs
LucaCoduriV: Yes we can of course, but I made it this way so that it's easier to read the code.
Do you want maybe to make a sub module instead ?

module name would be "handle_window_moved_or_resized_end"
in mod.rs I would put the "handle_window_moved_or_resized_end" function
and the two others file in this module.

What do you think about that ?

packages/wm/src/common/events/handle_window_moved_end.rs
LucaCoduriV: Yep, I saw that function but I don't know why I didn't used it...
But it's done now :)

packages/wm/src/common/events/handle_window_moved_end.rs
LucaCoduriV: Wow nice that simplifies it a lot ! Thanks !

done

packages/wm/src/common/events/handle_window_moved_or_resized_end.rs
lars-berger: even if it's left as separate, `window_moved_end` would be very long. since `handle_window_moved_or_resized_end` and `window_resized_end` are just a few lines, it doesn't add much complexity if they're grouped together and it would be consistent with the rest of the event handlers

packages/wm/src/common/events/handle_window_moved_end.rs
```diff
@@ -0,0 +1,323 @@
...
+    }
+    _ => {}
+  }
+  moved_window.set_window_operation(WindowOperation::default());
+  state
+    .pending_sync
+    .containers_to_redraw
+    .push(Container::Workspace(
+      window_under_cursor.workspace().unwrap(),
+    ));
```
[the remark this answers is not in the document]
LucaCoduriV: Nice feature, I haven't seen that

done

packages/wm/src/common/events/handle_window_moved_end.rs
```diff
@@ -0,0 +1,323 @@
...
+    .context("couldn't find the new tiled window")?
+    .as_tiling_window()
+    .context("window is not a tiled window")?
+    .clone();
+
+  // TODO: We can optimize that by not detaching and attaching the window
+  // Little trick to get the right index
+  detach_container(Container::TilingWindow(moved_window.clone()))?;
+  let target_window_index = target_window.index();
+  attach_container(
```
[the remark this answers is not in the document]
LucaCoduriV: No the reason is because if the moved window is in the same direction container as the target window, when getting the index directly it will be moved to the wrong place because everything will be shifted.
At least this is what was happening, but I'll try to remove it to see what happens now.

packages/wm/src/common/events/handle_window_moved_or_resized_end.rs
LucaCoduriV: Okay, I moved everything into the same file

done

packages/wm/src/common/events/handle_window_moved_end.rs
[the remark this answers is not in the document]
LucaCoduriV: So I tested to remove those lines and it's still happening. I either need to keep those, or to find an other solution.
What I like with my solution, is that it only takes 5 lines. But of course it might be slower than an other one.

I tried once to calculate the future position of the "target_window" if we remove the "moved_window", but without success.
But I didn't tried very long to be honest.

packages/wm/src/windows/window_operation.rs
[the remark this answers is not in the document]
LucaCoduriV: > i'm thinking the `ActiveDrag` struct would be wrapped in `Option` and only exists if the window is being dragged
> 
> ```rust
> pub trait WindowGetters {
>   fn active_drag(&self) -> Option<ActiveDrag>;
>   ...
> ```

Oops sorry havn't seen you comment, do you want me to change that ?

Issue #619, opened 2024-06-09
[Feature Request] Drag window to different monitor moves window to workspace


nathanautrique:
Hi,

First of all, I love the app!

The request: 
Since different monitors have different workspaces, it would be nice if dragging a window from one monitor to another could change it to that workspace. This could also be locked behind a shortcut, for example moving while holding ALT.

Context:
I'm mostly using my keyboard, but for some programs you just really have to use the mouse. 
In those cases it would be nice to keep my hand on the mouse and just drag the windows to another monitors workspace

lars-berger:
This has now been added 🎉 [Latest release](https://github.com/glzr-io/glazewm/releases)

PR #673, opened 2024-08-12
feat: be able to change tiling direction when moving window via mouse

### LINKED ISSUE #564: Move windows with mouse

LucaCoduriV:
closes #564

### Description
This pull request addresses two bugs related to window management within containers and one enhancement:

1. **Bug 1: Incomplete Window Reorganization**
   - **Issue**: When a window is inside a container, dragging the window sometimes leaves gaps in the tiling arrangement.
   - **Discovery**: This issue occurs because the container was not being added to the pending sync vector.
   - **Solution**: Add the container to the pending sync vector to ensure proper reorganization of windows.

2. **Bug 2: Workspace Crash with Split Containers**
   - **Issue**: When at least three windows are in a split container, moving a window (highlighted with a red arrow in the screenshot) into the child split causes an error stating that the parent has no workspace, leading to a crash.
   - **Discovery**: The crash happens when attempting to move the window into the workspace before setting it as a floating window.
   - **Solution**: Fix the issue by ensuring the dragged window is moved into the workspace before it is set as a floating window.
image
3. **Enhancement: Improved Window Splitting Options**
   - **Feature**: Previously, window splitting was limited in direction. This enhancement allows users to split windows in all four directions: top, bottom, left, and right.
   - **Benefit**: Provides greater flexibility in window management, allowing for more customized layouts based on user preferences.


lars-berger:
Closes #619

PR #676, opened 2024-08-15
docs: typo and grammar fixes in readme

tmsagarofficial:
Just some minor typo/grammar fixes to allow better readability and understanding

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,7 +272,7 @@ window_behavior:
   initial_state: "tiling"
 
   # Sets the default options for when a new window is created. This also
-  # changes the defaults for when the state changes commands, like
+  # changes the defaults for when the state change commands, like
   # `set-floating`, is used without any flags.
   state_defaults:
     floating:
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ window_behavior:
 
     fullscreen:
       # Maximize the window if possible. If the window doesn't have a
-      # maximize button, then it'll be fullscreen normally instead.
+      # maximize button, then it'll be made fullscreen normally instead.
       maximized: false
 ```
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ GlazeWM lets you easily organize windows and adjust their layout on the fly by u
 
 ## Installation
 
-**The latest version of GlazeWM is downloadable via [releases](https://github.com/glzr-io/GlazeWM/releases).** Zebar can optionally be installed as well via a checkbox in the install.
+**The latest version of GlazeWM is downloadable via [releases](https://github.com/glzr-io/GlazeWM/releases).** Zebar can optionally be installed as well via a checkbox during installation.
 
 Scoop and Winget are currently not supported. If you instead want to build from source, refer to the
 [contributing guide](https://github.com/glzr-io/glazewm/blob/main/CONTRIBUTING.md).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -272,8 +272,8 @@ window_behavior:
   initial_state: "tiling"
 
   # Sets the default options for when a new window is created. This also
-  # changes the defaults for when the state change commands, like
-  # `set-floating`, are used without any flags.
+  # changes the defaults for when the state changes commands, like
```
lars-berger, line 275: `state change commands` refers to commands that can change state

README.md
```diff
@@ -284,7 +284,7 @@ window_behavior:
 
     fullscreen:
       # Maximize the window if possible. If the window doesn't have a
-      # maximize button, then it'll be fullscreen'ed normally instead.
+      # maximize button, then it'll be fullscreen normally instead.
```
lars-berger, line 287: sounds a bit strange - shouldn't fullscreen be past participle? e.g. made fullscreen

tmsagarofficial: Ohh ok , I also felt this dilemma of which is the right usage while trying to find references on the internet.

But I think we can also use other alternatives like:
`If maximizing isn't possible, the window will default to fullscreen mode.`
`When there's no maximize button, fullscreen mode will be used by default.`

@lars-berger what do u think suits the context well?

lars-berger: Both are fullscreen states (ie fullscreen with maximize and fullscreen without maximize). We don't really have a good term for separating the two - 'fullscreen maximized' and what's called just 'normal fullscreen' for now

tmsagarofficial: 🤔 Hmmm, makes sense .

I'll revert it to what it was before then?

lars-berger: we could do `made fullscreen normally` instead?

tmsagarofficial: Yeah tht LGTM😅

README.md
```diff
@@ -71,7 +71,7 @@ general:
   focus_follows_cursor: false
 
   # Whether to switch back and forth between the previously focused
-  # workspace when focusing the current workspace.
+  # workspace when focusing on the current workspace.
```
lars-berger, line 74: `focusing the current workspace` sounds more correct imo

README.md
```diff
@@ -273,7 +273,7 @@ window_behavior:
 
   # Sets the default options for when a new window is created. This also
   # changes the defaults for when the state change commands, like
-  # `set-floating`, are used without any flags.
+  # `set-floating`, is used without any flags.
```
lars-berger, line 276: should be reverted to `are`

lars-berger:
Couple final comments, otherwise lgtm

README.md
tmsagarofficial: I think without using the `on` then it would convey the meaning like it is some other process/action of focussing the current workspace. but with `on` used the message conveyed means that the focus can be changed from the current to the previously focussed workspace.

What do u say?

lars-berger: mm it does make sense. focusing (without `on`) is more consistent with the language we use elsewhere though

README.md
tmsagarofficial: 👍🏼

tmsagarofficial:
check now @lars-berger

README.md
lars-berger: should be `made fullscreen normally`

lars-berger:
Sweet! thanks for the help on this 🙌

PR #683, opened 2024-08-19
feat: Focus command to focus on specific monitor by index

hiszd:
This PR is to add a command option to the `focus` command that allows you to focus on a specific monitor by index.
And example usage of this would be:
```
glazewm.exe command focus --monitor 1
```
This would focus on the second monitor.
The indexing is zero-based, so the first monitor would be monitor 0.

### DIFF
--- a/packages/wm/src/monitors/commands/focus_monitor.rs
+++ b/packages/wm/src/monitors/commands/focus_monitor.rs
@@ -1,5 +1,6 @@
+use anyhow::Context;
+
 use crate::{
-  containers::traits::CommonGetters,
   user_config::UserConfig,
   wm_state::WmState,
   workspaces::{commands::focus_workspace, WorkspaceTarget},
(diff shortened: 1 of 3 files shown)


packages/wm/src/monitors/commands/focus_monitor.rs
```diff
@@ -0,0 +1,81 @@
...
+  let focused_workspace = state
+    .focused_container()
+    .and_then(|focused| focused.workspace())
+    .context("No workspace is currently focused.")?;
+
+  let monitors = state.monitors();
+
+  // if there are fewer monitors than the index provided error out and bail
+  // early from function
+  if monitors.len() < target + 1 {
```
lars-berger, line 26: could do `monitors.get(target_index)` instead. otherwise this allows for negative indices

packages/wm/src/monitors/commands/focus_monitor.rs
```diff
@@ -0,0 +1,81 @@
...
+
+use crate::{
+  containers::{commands::set_focused_descendant, traits::CommonGetters},
+  wm_state::WmState,
+  workspaces::commands::deactivate_workspace,
+};
+
+/// Focuses a monitor by a given monitor index.
+///
+/// The monitor will be activated if it isn't already active.
```
lars-berger, line 12: remove this bottom comment - doesn't really make sense

packages/wm/src/monitors/commands/focus_monitor.rs
```diff
@@ -0,0 +1,81 @@
...
+  // if there are fewer monitors than the index provided error out and bail
+  // early from function
+  if monitors.len() < target + 1 {
+    anyhow::bail!("target index greater than number of monitors");
+  }
+
+  let target_monitor = monitors[*target].clone();
+
+  if target_monitor.has_focus(None) {
+    anyhow::bail!("target monitor already has focus");
```
lars-berger, line 33: no need to bail, can no-op instead. bailing will cause an error dialog to show up when the user runs this command

packages/wm/src/monitors/commands/focus_monitor.rs
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  if let Some(target_workspace) = target_monitor.displayed_workspace() {
```
lars-berger, line 36: can simplify this block by using the existing `focus_workspace` fn, e.g. `focus_workspace(WorkspaceTarget::name(displayed_workspace.config().name))`

lars-berger (changes requested):

lars-berger:
Sweet, thanks for the pr! Some minor fixes otherwise it's looking good

lars-berger:
Thank you 🙌 Hope you don't mind but did some minor tweaks so that we could avoid the `unwrap` calls

PR #696, opened 2024-08-27
feat: add support for using env variable `GLAZEWM_CONFIG_PATH` for config path

### LINKED ISSUE #689: [Feature Request] Specify config file with an env variable

All other apps i use allow me to customize the location of the config file, it's nice having all logic (so location for "where should config files be located") in a single place 

If `GLAZEWM_CONFIG_FILE` env variable is set, use that as the config file.

### Alternatives considered

_No response_

nikitarevenco:
Closes #689

### DIFF
--- a/packages/wm/src/user_config.rs
+++ b/packages/wm/src/user_config.rs
@@ -1,4 +1,5 @@
 use std::{collections::HashMap, fs, env, path::PathBuf};
+
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 
--- a/packages/wm/src/user_config.rs
+++ b/packages/wm/src/user_config.rs
@@ -1,5 +1,4 @@
-use std::{collections::HashMap, fs, path::PathBuf};
-
+use std::{collections::HashMap, fs, env, path::PathBuf};
 use anyhow::{Context, Result};
 use serde::{Deserialize, Serialize};
 
--- a/packages/wm/src/user_config.rs
+++ b/packages/wm/src/user_config.rs
@@ -42,7 +42,12 @@ impl UserConfig {
       .context("Unable to get home directory.")?
       .join(".glzr/glazewm/config.yaml");
 
-    let config_path = config_path.unwrap_or(default_config_path);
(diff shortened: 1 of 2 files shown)


packages/wm/src/user_config.rs
```diff
@@ -42,7 +42,12 @@ impl UserConfig {
       .context("Unable to get home directory.")?
       .join(".glzr/glazewm/config.yaml");
 
-    let config_path = config_path.unwrap_or(default_config_path);
+    let env_config_path = env::var("GLAZEWM_CONFIG_FILE").ok().map(PathBuf::from);
```
lars-berger, line 45: could we rename this to `GLAZEWM_CONFIG_PATH`?

packages/wm/src/user_config.rs
```diff
@@ -42,7 +42,12 @@ impl UserConfig {
...
+
+    let config_path = config_path
+        .or(env_config_path)
```
lars-berger, line 48: we could use `or_else` here so that the env var doesn't have to be unnecessarily evaluated

```rust
    let config_path = config_path
      .or_else(|| env::var("GLAZEWM_CONFIG_PATH").ok().map(PathBuf::from))
      .unwrap_or(default_config_path);
```

lars-berger (changes requested):

lars-berger:
Sweet, thanks for the PR 🙌

packages/wm/src/user_config.rs
nikitarevenco: Ok, yes that's a good idea! i added it

packages/wm/src/user_config.rs
nikitarevenco: Sure

nikitarevenco:
> Sweet, thanks for the PR 🙌

Note: Instead of specifying `GLAZEWM_CONFIG_PATH`, which will state the *directory* in which the config file is found, how about we have `GLAZEWM_CONFIG_FILE` which will 

> Sweet, thanks for the PR 🙌

I also added a note to the readme on the new env variable

lars-berger:
Perfect thank you!

PR #702, opened 2024-08-29
feat: add command `set-tiling-direction [vertical|horizontal]`

JonasWischeropp:
I added the two commands `set-tiling-direction-vertical` and `set-tiling-direction-horizontal`
in order to be able to set the tiling direction faster and in the same way it is done in `i3wm`.
When using `toggle-tiling-direction` one always has to identify the current direction first.

Let me know what you think. Specially whether using `panic!` here is the right method:
https://github.com/JonasWischeropp/glazewm/blob/3c3fd07cd667c7e6165aa422d144e88ee0858af0/packages/wm/src/containers/commands/toggle_tiling_direction.rs#L110

Thank you for this amazing window manager.

### DIFF
--- a/packages/wm/src/app_command.rs
+++ b/packages/wm/src/app_command.rs
@@ -227,8 +227,10 @@ pub enum InvokeCommand {
   ToggleMinimized,
   ToggleTiling,
   ToggleTilingDirection,
-  SetTilingDirectionVertical,
-  SetTilingDirectionHorizontal,
+  SetTilingDirection {
+    #[clap(required = true)]
+    tiling_direction: TilingDirection,
+  },
   WmCycleFocus {
     #[clap(long, default_value_t = false)]
     omit_fullscreen: bool,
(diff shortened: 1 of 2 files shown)


packages/wm/src/containers/commands/toggle_tiling_direction.rs
```diff
@@ -82,3 +83,39 @@ fn toggle_window_direction(
...
+  direction: TilingDirection,
+) -> anyhow::Result<()> {
+  let current_direction = tiling_direction(&container)?;
+  if current_direction != direction {
+    toggle_tiling_direction(container, state, config)
+  }
+  else {
+    Ok(())
+  }
+}
```
lars-berger, line 100: can be simplified to:

```rust
  let direction_container = container
    .direction_container()
    .context("No direction container.")?;

  match direction_container.tiling_direction() == tiling_direction {
    true => Ok(()),
    false => toggle_tiling_direction(container, state, config)
  }
```

packages/wm/src/containers/commands/toggle_tiling_direction.rs
```diff
@@ -82,3 +83,39 @@ fn toggle_window_direction(
 
   Ok(split_container.into())
 }
+
+pub fn set_tiling_direction(
+  container: Container,
+  state: &mut WmState,
+  config: &UserConfig,
+  direction: TilingDirection,
```
lars-berger, line 91: rename `direction` -> `tiling_direction`

packages/wm/src/app_command.rs
```diff
@@ -225,6 +227,8 @@ pub enum InvokeCommand {
   ToggleMinimized,
   ToggleTiling,
   ToggleTilingDirection,
+  SetTilingDirectionVertical,
```
lars-berger, line 230: create single `set-tiling-direction <TILING_DIRECTION>` command instead of two separate

lars-berger (changes requested):

lars-berger:
Thanks for the pr 🙌 was looking to get this added at some point

Added some minor fixups otherwise lgtm

JonasWischeropp:
Thank you for the feedback. I have adjusted the code acordingly.

lars-berger:
Awesome, thank you ❤️

Issue #653, opened 2024-08-06
Disable border highlight for the focused window on a specific window.


a0w89fwa0f:
Is it possible to disable the focused border highlight for a specific window.

I use an extension for Visual Studio Code which makes the window transparent and the border highlight color fills the entire window making it unusable: https://github.com/illixion/vscode-vibrancy-continued

Thanks

lars-berger:
We could add a `match` option like like with window rules, e.g.
```yaml
window_effects:
  other_windows:
    border:
      enabled: true
      color: '#a1a1a1'
      match:
        - window_process: { equals: 'Code' }
```

JonasWischeropp:
The current version is now fully functional and is using `window_effects`.
Furthermore I have added the ability to change the corner style of windows.

```yaml
focused_window:
  hide_title_bar:
    enabled: true
  corner:
    enabled: true
    # Allowed values: 'windows_default', 'square', 'round', 'small_round'
    style: 'square'
```

But I have noticed that removing the title bar of some windows can cause problems, e.g:
- notepad: visuals of the title bar are wrong
- windows file explorer: tabs in the title bar no longer fully work

Therefore it would be nice to have the ability to `match` like with `window_rules` (like discussed in #653).

But I think it could be beneficial to use `window_rules` instead of `window_effects`.
This would require the introduction of `WindowRuleEvent::Unfocus`.
For example instead of using:
```yaml
window_effects:
  focused_window:
    border:
      enabled: true
      color: '#8dbcff'
    hide_title_bar:
      enabled: false

  other_windows:
    border:
      enabled: true
      color: '#a1a1a1'
    hide_title_bar:
      enabled: true
```
One could use:
```yaml
window_rules:
  - commands: ['set-title-bar-visibility false', 'set-border-color #8dbcf']
  # (set-border-color would need to be implemented)
    on: 'focus'
    run_once: false
    match:
      - window_class: { equals: 'example' }

  - commands: ['set-title-bar-visibility true', 'set-border-color #a1a1a1']
    on: 'unfocus'
    run_once: false
    match:
      - window_class: { equals: 'example' }
```
(Could be made cleaner by making `match` optional and matching everything when it is not provided.)

I think this system would be more extendable.
It would be easier to introduce new rules what effects should be applied when.
Because it allows to change effects not only when the focus changes but also on other `WindowRuleEvent`s.
Currently there are only three but others could be exposed in the future.
[...]

PR #735, opened 2024-09-21
feat: add window effects for hiding title bar and changing corner style 

### LINKED ISSUE #504: [Feature Request] Remove title bars from windows
I think most people that use a wm dont like these annoying white bars, but you cant really turn them of, or can you (without just removing all decorations icluding the rounded corners)? I managed to write some Odin code to get rid of the on the few apps that I use that actually have them, however it would be really convenient to have this be feature of glaze, wich is a really good wm btw, big thanks to everyone who participates in this project. Anyway here is my code, you could probably use it as a starting point for a c# implemenation, or maybe this at least helps someone out. I would implement this myself, but Im just not good at c#.
Further reading:
 - https://learn.microsoft.com/en-us/windows/apps/desktop/modernize/apply-rounded-corners
 - https://learn.microsoft.com/en-us/windows/win32/winmsg/window-styles
```go
package title_remover

import "core:fmt"
import "core:os"
import "core:runtime"
import "core:slice"
import "core:strings"

import win "core:sys/windows"

// if any title _CONTAINS_ one of these strings it will get its borders removed
MANAGE :: []string{"CactusViewer", "Minecraft"}

enum_windows_proc :: proc "stdcall" (hWnd: win.HWND, lParam: win.LPARAM) -> win.BOOL {
	context = runtime.default_context()

	handles := transmute(^[dynamic]win.HWND)lParam
	append(handles, hWnd)

	return true
}

main :: proc() {
	handles := make([dynamic]win.HWND)
	win.EnumWindows(enum_windows_proc, transmute(win.LPARAM)&handles)

	for handle in handles {
		str := make([]u16, 1024)

		n := win.GetWindowTextW(handle, raw_data(str), 1023)

		manageable := true

		title, _ := win.utf16_to_utf8(str[:n])

		if manageable {
			for m in MANAGE {
				if strings.contains(title, m) {
					remove_window_title_bar(handle)
					break
				}
			}
		}

	}
}

remove_window_title_bar :: proc(hWnd: win.HWND) -> bool {
	if hWnd == nil {
		return false
	}

	preference: i32 = 2
	result := win.DwmSetWindowAttribute(
		hWnd,
		auto_cast win.DWMWINDOWATTRIBUTE.DWMWA_WINDOW_CORNER_PREFERENCE,
		&preference,
		size_of(preference),
	)

	return(
		win.SUCCEEDED(result) &&
		win.SetWindowLongW(hWnd, win.GWL_STYLE, transmute(i32)win.WS_POPUPWINDOW) != 0 \
	)
}
```
PS:  command: "resize borders 0px -7px -7px -7px", as a window rule in your config fixes the size of most windows without titlebars

JonasWischeropp:
This PR adds the command `set-title-bar-visibility <true|false>` to enable hiding title bars.
This closes #504.

For example all title bars can then be deactivated with:
```yaml
window_rules:
  - commands: ['set-title-bar-visibility false']
    match:
      - window_class: { regex: '.*' }
```
Windows that have custom titlebars are uneffected.


lars-berger:
Ay this is great, thanks for the pr 🙌 

Reckon it'd be best to instead add it as a window effect like we've got with window borders, e.g.

```yaml
window_effects:
  focused_window:
    hide_title_bar:
      enabled: true

  other_windows:
    hide_title_bar:
      enabled: true
```

If you're down to make the change, modify the `apply_window_effects` fn in `platform_sync`

JonasWischeropp:
I will look into it.

JonasWischeropp:
The current version is now functional and is using `window_effects`.
Furthermore I have added the ability to change the corner style of windows.

```yaml
focused_window:
  hide_title_bar:
    enabled: true
  corner:
    enabled: true
    # Allowed values: 'windows_default', 'square', 'round', 'small_round'
    style: 'square'
```

But I have noticed that removing the title bar of some windows can cause problems, e.g:
- notepad: visuals of the title bar are wrong
- windows file explorer: tabs in the title bar no longer fully work

Therefore it would be nice to have the ability to `match` like with `window_rules` (like discussed in #653).

But I think it could be beneficial to use `window_rules` instead of `window_effects`.
This would require the introduction of `WindowRuleEvent::Unfocus`.
For example instead of using:
```yaml
window_effects:
  focused_window:
    border:
      enabled: true
      color: '#8dbcff'
    hide_title_bar:
      enabled: false

  other_windows:
    border:
      enabled: true
      color: '#a1a1a1'
    hide_title_bar:
      enabled: true
```
One could use:
```yaml
window_rules:
  - commands: ['set-title-bar-visibility true', 'set-border-color #8dbcf']
  # (set-border-color would need to be implemented)
    on: 'focus'
    run_once: false
    match:
      - window_class: { equals: 'example' }

  - commands: ['set-title-bar-visibility false', 'set-border-color #a1a1a1']
    on: 'unfocus'
    run_once: false
    match:
      - window_class: { equals: 'example' }
```
(Could be made cleaner by making `match` optional and matching everything when it is not provided.)

I think this system would be more extendable.
It would be easier to introduce new rules what effects should be applied when.
Because it allows to change effects not only when the focus changes but also on other `WindowRuleEvent`s.
Currently there are only three but others could be exposed in the future.
For example with a `WindowRuleEvent` like `BecameFloating` one could have round corners in floating mode and square corners in tiling mode.
This would probably also require new `MatchType`s but what I am trying to say is that I think that using `window_rules` would be more flexible in the future.


What do you think?

JonasWischeropp:
Matching in `window_effects` can also only enable/disable the effect but not apply it differently to windows. For example different border colors for different windows.

lars-berger:
Will merge as is but will likely do some naming tweaks before release. There's a bug with the border color not being set reliably, so gonna do some changes to how window effects work after the merge

> But I think it could be beneficial to use window_rules instead of window_effects.
> This would require the introduction of WindowRuleEvent::Unfocus.

True that's a good point. If we added an `unfocus` events, window effects could essentially be entirely implemented via window rules. The big upside with having a separate window effects config imo though is that it's a lot simpler to configure, and someone wouldn't have to figure how commands and window rule triggers work. Ideally it should be as simple as possible - e.g. just flip a boolean for whether to remove title bars. Still think that a window effects config has its place but would absolutely be down to add an `unfocus` window rule event 👍

lars-berger:
> There was a bug where the win11 window animation would disappear after disable the titlebar .
> 
> refer to :
> 
> [ptazithos/glazewm-extra#11](https://github.com/ptazithos/glazewm-extra/issues/11)

Not sure whether we'd be able to change that behavior unfortunately. The animations are built into the OS and we don't do anything on our side to trigger the animation

PR #748, opened 2024-09-29
ci: add separate winget release workflow

vikingnope:
Changed winget workflow to correctly run as intended during publish of new release

### DIFF
--- a/.github/workflows/winget-release.yaml
+++ b/.github/workflows/winget-release.yaml
@@ -8,7 +8,7 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@v2
+      - uses: vedantmgoyal2009/winget-releaser@main
         with:
           identifier: glzr-io.glazewm
           installers-regex: 'glazewm-[0-9.]+\.exe$'
--- a/.github/workflows/winget-release.yaml
+++ b/.github/workflows/winget-release.yaml
@@ -8,7 +8,7 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-      - uses: vedantmgoyal2009/winget-releaser@main
+      - uses: vedantmgoyal2009/winget-releaser@b87a066d9e624db1394edcd947f8c4e5a7e30cd7
         with:
           identifier: glzr-io.glazewm
           installers-regex: 'glazewm-[0-9.]+\.exe$'
--- a/.github/workflows/winget-release.yaml
+++ b/.github/workflows/winget-release.yaml
@@ -1,7 +1,7 @@
 name: Winget Release
 on:
   release:
-    types: [released]
+    types: [published]
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


lars-berger:
It errors because the workflow has an option to publish as draft, which then causes the winget release step to not find it :(

image

Regex and version identifiers should actually be working fine. Think the best solution here would be to move out the winget release to a separate workflow triggered on release, e.g.

```yaml
name: Winget publish
on:
  release:
    types: [published]
```

vikingnope:
Oh yep you are right I didn't notice that, let me do it

.github/workflows/winget-release.yaml
```diff
@@ -0,0 +1,14 @@
+name: Winget Release
+on:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@v2
```
lars-berger, line 10: pin this to latest commit `vedantmgoyal2009/winget-releaser@b87a066d9e624db1394edcd947f8c4e5a7e30cd7`

.github/workflows/winget-release.yaml
```diff
@@ -0,0 +1,14 @@
...
+        with:
+          identifier: glzr-io.glazewm
+          installers-regex: 'glazewm-[0-9.]+\.exe$'
+          token: ${{ secrets.WINGET_TOKEN }}
```
lars-berger, line 14: nit: add newline at end of file

.github/workflows/winget-release.yaml
```diff
@@ -0,0 +1,14 @@
+name: Winget Release
```
lars-berger, line 1: nit: `Winget release` for same casing as other workflows

.github/workflows/winget-release.yaml
```diff
@@ -0,0 +1,14 @@
...
+on:
```
lars-berger, line 2: add dispatch trigger so that it can be manually run (like to get the current release added)
```yaml
on:
  workflow_dispatch:
  release:
   ...
```

lars-berger:
Ay sweet, thanks for taking care of this. Just some small style changes otherwise lgtm

.github/workflows/winget-release.yaml
```diff
@@ -0,0 +1,15 @@
...
+  workflow_dispatch:
+  release:
+    types: [published]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: vedantmgoyal2009/winget-releaser@main
```
lars-berger, line 11: might be a bit over-cautious but `@main` can be a security risk since the author is able to update it to something malicious whenever. so you'll often see ppl pinning it like so `vedantmgoyal2009/winget-releaser@b87a066d9e624db1394edcd947f8c4e5a7e30cd7`

lars-berger:
Thank you 🙌

vikingnope:
If you try to run it right now it won't work as I already opened a PR for it

lars-berger:
Absolute chad

PR #756, opened 2024-10-01
feat: scale inner and outer gaps by monitor DPI 

JonasWischeropp:
I was trying to make my zebar config work with multiple monitors that have different display scaling factors.
Related discussion: glzr-io/zebar#72
This PR suggests a possible fix by scaling `outer_gap` and `inner_gap` by the display scale.
Let me know if I should change something.

### DIFF
--- a/packages/wm/src/containers/traits/tiling_size_getters.rs
+++ b/packages/wm/src/containers/traits/tiling_size_getters.rs
@@ -77,7 +77,14 @@ macro_rules! impl_tiling_size_getters {
       }
 
       fn inner_gap(&self) -> LengthValue {
-        self.0.borrow().inner_gap.clone()
+        let scale = match self.monitor() {
+          None => 1_f32,
+          Some(monitor) => match monitor.native().dpi() {
+            Ok(dpi) => dpi,
+            Err(_) => 1_f32,
+          },
+        };
+        self.0.borrow().inner_gap.clone() * scale
       }
 
       fn set_inner_gap(&self, inner_gap: LengthValue) {
--- a/packages/wm/src/common/length_value.rs
+++ b/packages/wm/src/common/length_value.rs
@@ -1,4 +1,4 @@
-use std::str::FromStr;
+use std::{ops::Mul, str::FromStr};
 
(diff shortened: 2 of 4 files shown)


packages/wm/src/containers/traits/tiling_size_getters.rs
```diff
@@ -77,7 +77,14 @@ macro_rules! impl_tiling_size_getters {
...
       fn inner_gap(&self) -> LengthValue {
-        self.0.borrow().inner_gap.clone()
+        let scale = match self.monitor() {
+          None => 1_f32,
+          Some(monitor) => match monitor.native().dpi() {
+            Ok(dpi) => dpi,
+            Err(_) => 1_f32,
+          },
+        };
+        self.0.borrow().inner_gap.clone() * scale
```
lars-berger, line 87: like in the Zebar PR, think we should only be scaling pixels here

would suggest adding a new field:
```rust
#[derive(Clone, Debug, Deserialize, Serialize)]
#[serde(rename_all(serialize = "camelCase"))]
pub struct GapsConfig {
  /// Whether to scale the gaps with the DPI of the monitor.
  #[serde(default = "default_bool::<true>")]
  pub scale_with_dpi: bool,
  
// sample-config.yaml:
// gaps:
//   # Whether to scale the gaps with the DPI of the monitor.
//   scale_with_dpi: true
```
that way, we can make it more explicit that this gets scaled, since all other px values in the config deal with physical pixels.

would then have to change `TilingSizeGetters::inner_gap`, `TilingSizeGetters::set_inner_gap`, `Workspace::set_outer_gap` to instead get/set the `GapsConfig` struct. 

to deal with only scaling px values and not %, maybe we could add something like `LengthValue::to_px_scaled(&self, total_px: i32, scale_factor: f32)`?

JonasWischeropp: Sounds good.

lars-berger:
Will merge as is and add the `gaps.scale_with_dpi` config option in a follow-up PR 👍

JonasWischeropp:
Ok, thank you.

PR #758, opened 2024-10-01
ci: fix regex for detecting latest version in Winget release workflow

vikingnope:
As said above, the missing 'v' stops from finding the url


lars-berger:
❤️

Issue #250, opened 2023-02-13
[Bug] Microsoft Excel is loosing menu and ribbon bar


s-v-o:
I have a strange behaviour with Microsoft Excel.
  
When I start Excel, everything is fine: 
![image](https://user-images.githubusercontent.com/64085779/218489672-051c14dc-6047-4e3f-9e9f-3d93746ce68f.png)
  
After changing the workspace and comming back the menu and ribbon bar disappear:
![image](https://user-images.githubusercontent.com/64085779/218489964-97348b8d-8b9e-4b93-bce5-8c6ea1ef4f52.png)
  
Even stranger: Everything works fine in Word.
  
I cannot say excactly in which version of GlazeWM I first noticed this.

There are no new entries in errors.log.

CalvinTrop:
Same issue here, which unfortunately renders the project unusable for my current workflow...

kirpy0x:
Same here on Win11.  If I remember correctly, this didn't happen on Win10.

I thought it was an issue with my Excel Installation but then noticed they all disappeared right once a Window management action was done.

rwijtvliet:
Same here. FWIW, I'm on Win10 enterprise edition.

hongyuanjia:
Same here. I'm on Win 10 Pro 22H2. Word and PowerPoint work perfectly fine.

AttackOnTyler:
Can confirm, Toggling full screen doesn't remedy the issue either.

s-v-o:
FWIW: As a workaround I've added this to my config.yaml:
```
- command: "ignore"
  match_title: "/Excel/"
```
So Excel is ignored from GlazeWM and I could use it. But of course this way Excel is not managed by GlazeWM.

n-crespo:
Here to add another barely useful nugget of information. I'm having this exact problem, here's how I'm forcing it to work for my workflow. 

I have a few apps that I want Glaze to ignore, (they will appear in every workspace, keymaps don't work on them, etc) and one of those is PowerToys, because I use PowerToys Run to launch applications easily  (With the exception of my browsers/terminal/Spotify which I launch with Glaze keymaps) and I don't want Glaze to get confused by the popups and other tools.

[...]

PR #792, opened 2024-10-15
feat: hide and show windows via set_cloak from COM lib

### LINKED ISSUE #416: [Feature request] Switching workspace automatically when clicking the app icon in the system taskbar instead of creating a new instance
Hello. First of all, thank you for the great WM that delivers the best user experience among all the WM for Windows I have tried.

I have noticed that the Windows taskbar (not the bar generated by GlazeWM) only shows the icons of the apps that are opened in the current workspace. Therefore, if I have a browser window open in Workspace 1, and I am in Workspace 2, I am not able to switch back to the browser window by clicking the app icon in the taskbar. If the browser icon is pinned to the taskbar, clicking it while in Workspace 2 will create another instance of the app in Workspace 2. I don't think this is the expected behavior for most users. Would it be possible to show all the opened apps in the system taskbar and switch to the corresponding workspace that includes the target app when users click on the icon in the system bar?

rpop0:
Previous method of hiding the windows (SW_HIDE and SWP_HIDEWINDOW) would completely remove the window from the taskbar, which would lead to unexpected behavior for some people. set_cloak allows the hidden windows to still be present in the task bar.

Other discussions on the topic: https://github.com/Ciantic/AltTabAccessor/issues/1

This PR also would close #416


rpop0:
Draft, looking into having the cloaking functionality to be toggled in the config, allowing users to switch between cloaking and the previously used methods.

rmanthorpe:
@rpop0 I've been running this branch this morning and it also seems to fix https://github.com/glzr-io/glazewm/issues/250,  thank you so much!

rmanthorpe:
> Draft, looking into having the cloaking functionality to be toggled in the config, allowing users to switch between cloaking and the previously used methods.

Can we do this on a per-application basis? I suspect we will find that some applications (like Excel) only work with this method whereas others work better with the original.

I'm finding that MS Outlook doesn't have well with this method. Correctly belongs to a workspace but it won't tile, it's always full screen, even covering the zebar no matter what else I put in the same workspace.

rpop0:
> > Draft, looking into having the cloaking functionality to be toggled in the config, allowing users to switch between cloaking and the previously used methods.
> 
> 
> 
> Can we do this on a per-application basis? I suspect we will find that some applications (like Excel) only work with this method whereas others work better with the original.
> 
> 
> 
> I'm finding that MS Outlook doesn't have well with this method. Correctly belongs to a workspace but it won't tile, it's always full screen, even covering the zebar no matter what else I put in the same workspace.

Hm, from what I quickly tested now it seems like Outlook just crashes  the window manager when being started, leading to it not tiling. I'll investigate it and see why that happens.

rmanthorpe:
> it seems like Outlook just crashes the window manager when being started

You're quite right, it does indeed. I already had outlook open when I started glaze which led to the behaviour I saw.

rpop0:
@rmanthorpe I think I may have found a fix.

Can you try adding the following rules under the ignore command and see if the crash still happens?

```
- window_process: { equals: 'explorer' }
  window_class: { equals: 'ApplicationFrameWindow'}
```

rmanthorpe:
> Can you try adding the following rules under the ignore command and see if the crash still happens?

It still crashes I'm afraid. What I'm seeing (with or without that rule) is that opening outlook with glaze running crashes glaze but starting glaze with outlook running is fine and outlook is working perfectly normally. I'm not sure what happened the first couple of times where outlook wouldn't tile, I can't reproduce that behaviour now!

rpop0:
> > Can you try adding the following rules under the ignore command and see if the crash still happens?
> 
> 
> 
> It still crashes I'm afraid. What I'm seeing (with or without that rule) is that opening outlook with glaze running crashes glaze but starting glaze with outlook running is fine and outlook is working perfectly normally. I'm not sure what happened the first couple of times where outlook wouldn't tile, I can't reproduce that behaviour now!

So the ignore rules I provided above fix another crash, related to the login screen which MS uses for their apps (Login popup to sign in to Microsoft)

I think I narrowed down the issue to the splash screen shown when loading Outlook. WinLister also happens to crash when starting outlook.

Try adding the following rule to ignore, separate from the previous rules:
`- window_class: { equals: 'MsoSplash' }`

rmanthorpe:
> Try adding the following rule to ignore, separate from the previous rules:
> `- window_class: { equals: 'MsoSplash' }`

Yes! That has done the trick! This change is amazing is has had such a big impact on me today. I'm going to keep using this branch until it is merged.

rpop0:
@rmanthorpe The most recent commit should make the window manager more resistant to windows with weird behavior that would previously crash. While the rules I specified are longer really needed, they still provide a nicer experience by just ignoring those problematic windows.

rmanthorpe:
> The most recent commit should make the window manager more resistant to windows with weird behavior that would previously crash.

Tested and confirmed -  removed the config and Outlook is working well without it!

DreamMaoMao:
Any questions about this pr merger, which I think is working very well

rmanthorpe:
> Any questions about this pr merge? I think he's doing a really good job.

I've been running this branch (rebased onto master) for over two weeks now. It fixes several problems for me and I've had no issues since the outlook problem was fixed. I'd love to see it get over the line - what steps are necessary to get there?

rpop0:
> > Any questions about this pr merge? I think he's doing a really good job.
> 
> 
> 
> I've been running this branch (rebased onto master) for over two weeks now. It fixes several problems for me and I've had no issues since the outlook problem was fixed. I'd love to see it get over the line - what steps are necessary to get there?

I just didn't get the chance to work on it more, but i've been using it for this time too with no issues.

Mainly what needs to be done more is a config option to toggle between this cloaking minimization and the previous method of doing it. I'll try to find the time to get it done this weekend

PR #798, opened 2024-10-18
feat: add `not_equals` and `not_regex` match types

### LINKED ISSUE #786: [Bug] some window rules don't work
### Describe the bug

My window rule that sets some windows to floating don't apply to the Flow launcher settings and Steam windows that aren't the main one (except for the news window), the rule that ignores the window applies correctly to Steam Big Picture mode, but leaves a small gap at the bottom of the screen. This worked perfectly fine before updating to 3.4.0

### Reproduction

```yaml
window_rules:
  # Task Manager requires admin privileges to manage and should be ignored unless running
  # the WM as admin.
  - commands: ["ignore"]
    match:
      - window_process: { equals: "steamwebhelper" }
        window_title: { regex: ".*Big Picture.*" }

  - commands: ["set-floating"]
    match:
      - window_process: { equals: "steamwebhelper" }
        window_title: { regex: "(?!Steam).*" }

      - window_process: { equals: "Flow.Launcher" }
        window_title: { equals: "Settings" }

      - window_title: { equals: "Visual Studio Installer" }
```

_No response_

### Version number

3.4.0

RiceaRaul:
To @rpop0  "complaints" I added **not equals** for matchtype.

- Added a new match type: NotEquals, which checks for inequality.
- Extended the Regex match type to support an optional `not` parameter for negating regex matches.

An example of use would be for steam, you want the friend list and the chat to be ignored.
- window_title for friends list is Friends
- window_title for a chat is the person's name.
Normally you can use the regex method ^(?:(?!Steam).*$ but rust does not have support for Negative Lookahead.

Exemple of ussage:
- Not Equals:
```
- window_process: { equals: "steamwebhelper" }
  window_title: { not_equals: "Steam" }
```
-Not Regex
```
- window_process: { equals: "steamwebhelper" }
  window_title: { regex: "Steam", not: true }
```

At the same time, it fixes #786.

### DIFF
--- a/packages/wm/src/user_config.rs
+++ b/packages/wm/src/user_config.rs
@@ -596,9 +596,9 @@ pub struct WindowMatchConfig {
 pub enum MatchType {
   Equals { equals: String },
   Includes { includes: String },
-  Regex { regex: String},
+  Regex { regex: String },
   NotEquals { not_equals: String },
-  NotRegex {not_regex: String}
+  NotRegex { not_regex: String },
 }
 
 impl MatchType {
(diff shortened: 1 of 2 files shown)


lars-berger:
Ay this is a great idea 🙌 Thanks for submitting a PR

packages/wm/src/user_config.rs
```diff
@@ -605,9 +606,17 @@ impl MatchType {
     match self {
       MatchType::Equals { equals } => value == equals,
       MatchType::Includes { includes } => value.contains(includes),
-      MatchType::Regex { regex } => regex::Regex::new(regex)
-        .map(|re| re.is_match(value))
-        .unwrap_or(false),
+      MatchType::Regex { regex,not } => regex::Regex::new(regex)
```
lars-berger, line 609: would you be ok with changing the `not` option on the `regex` match type to be a separate `not_regex` instead? even though it's a bit worse of a name, I think it'd make it clearer to have consistency in the `not_*` naming (e.g. perhaps also `not_includes` in the future)

lars-berger:
Perfect thanks 👌 Will go ahead and merge

PR #801, opened 2024-10-20
feat: automatically deactivate non-current workspaces when all window…

DreamMaoMao:
Its feature is that when all Windows in a workspace (not necessarily the current one) are destroyed and that workspace is not currently being dispalyed, that workspace will also be automatically deactivate, making it more efficient to focus on the active workspace without having to focus on the windowless workspace again to deactivate it

https://github.com/user-attachments/assets/2a05b460-4cf1-4314-a03b-e906d0a0217c

### DIFF
--- a/packages/wm/src/common/events/handle_window_destroyed.rs
+++ b/packages/wm/src/common/events/handle_window_destroyed.rs
@@ -1,8 +1,10 @@
 use tracing::info;
 
 use crate::{
-  common::platform::NativeWindow, windows::commands::unmanage_window,
+  common::platform::NativeWindow, windows::commands::unmanage_window, 
   wm_state::WmState,
+  containers::traits::CommonGetters,
+  workspaces::commands::deactivate_workspace
 };
 
 pub fn handle_window_destroyed(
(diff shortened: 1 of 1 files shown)


packages/wm/src/common/events/handle_window_destroyed.rs
```diff
@@ -14,8 +16,16 @@ pub fn handle_window_destroyed(
   // Unmanage the window if it's currently managed.
   if let Some(window) = found_window {
     // TODO: Log window details.
+    let bind_workspace = window.workspace();
```
lars-berger, line 19: workspace should always be defined, so you can do
```rust
let workspace = window.workspace().context("No workspace.")?;
```
can then remove the `if let ...` conditional

packages/wm/src/common/events/handle_window_destroyed.rs
```diff
@@ -14,8 +16,16 @@ pub fn handle_window_destroyed(
...
+
     info!("Window closed");
     unmanage_window(window, state)?;
+
+    if let Some(workspace) = bind_workspace {
+      if !workspace.config().keep_alive && !workspace.has_children() && !workspace.is_displayed() {
```
lars-berger, line 25: add comment, eg. `// Destroy parent workspace if window was killed while its workspace was not displayed (e.g. via task manager).`

lars-berger:
Thanks for the PR! Wouldn't have thought to cover this edge case 🙌

packages/wm/src/common/events/handle_window_destroyed.rs
DreamMaoMao: Optimization completed

packages/wm/src/common/events/handle_window_destroyed.rs
DreamMaoMao: Optimization completed

PR #807, opened 2024-10-21
feat: Added wait command

CosminApo:
Added a command to wait for n seconds.

Example of usage:

```yaml
 startup_commands: [
    'shell-exec zebar',
    'focus --workspace 2',
    'shell-exec chrome.exe --new-window https://fakeurl.com',
    'wait 5',
    'focus --workspace 3',
    'shell-exec chrome.exe --new-window https://fakeurl.com',
  ]
```

### DIFF
--- a/packages/wm/src/common/events/handle_window_destroyed.rs
+++ b/packages/wm/src/common/events/handle_window_destroyed.rs
@@ -1,8 +1,11 @@
+use anyhow::Context;
 use tracing::info;
 
 use crate::{
   common::platform::NativeWindow, windows::commands::unmanage_window,
   wm_state::WmState,
+  containers::traits::CommonGetters,
+  workspaces::commands::deactivate_workspace
 };
 
 pub fn handle_window_destroyed(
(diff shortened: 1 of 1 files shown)


lars-berger:
Thanks for the PR @CosminApo!

One issue with sleeping the thread like this is that it'll fully block other events and commands from being processed while waiting. What's the use case you're looking to cover btw? Like eg with the `startup_commands` above, I think you could `shell-exec` the two Chrome windows and instead have window rules to move them to the correct workspace (based on title).

Either way, would still be open to add a command like this but it'd have to be non-blocking. One approach to achieve that is to have a channel with commands to be executed by the WM and spawn a tokio task that emits to the channel after X milliseconds. The channel (`tokio::mpsc::unbounded_channel`) could take a tuple of `InvokeCommand` and `Uuid` of the subject container, and could be added as a field to `WmState`. If you're down to take a stab at this lmk 👍 There's a [discord channel](https://discord.com/invite/ud6z3qjRvM) if you'd need help - feel free to dm me on there

CosminApo:
Thanks for the feedback @lars-berger . I'll have a play around with the implementation you suggested, but might be a while before I get to it. I'm closing this PR for now and I'll make a new one/re-open it once I get it working with the tokio stuff.

Issue #750, opened 2024-09-30
[Feature Request] Move all windows of current workspace to another at once


ssdu05:
### Describe the problem/motivation

Lets say you have 3 windows in workspace 1 and want to move all 3 windows to workspace 2. You'd have to move each window with shift + alt + <workspace_number> to workspace 2 and then jump back with alt + <workspace_number> to workspace 1. Repeat all of that for each window in workspace 1.

### Describe the solution you'd like

It would be faster and easier if there would be a keybind where you could move all windows of 1 workspace to another workspace. For example alt + q + <workspace_number>.

rmanthorpe:
Not quite the solution you're after but... if you bind `shift + alt + 1` to `move --workspace 1` etc instead of `move` & `focus` you can move and focus independently. That way you can move all the windows around the shift focus with `shit + alt + 1` when you're done.

gustaphe:
In i3wm, this is normally done by renaming the workspace (i.e. workspace numbers are just whatever number is first in the name). #776 could tie into this.

PR #810, opened 2024-10-23
feat: move all window to specific workspace

Muffeter:
#750 

This PR implements a new feature allowing users to move all windows to a specified workspace using custom key bindings. 

usage:
```
  - commands: ["move-all --workspace 1"]
    bindings: ["alt+q+1"]
```

### DIFF
--- a/packages/wm/src/app_command.rs
+++ b/packages/wm/src/app_command.rs
@@ -25,8 +25,9 @@ use crate::{
   user_config::{FloatingStateConfig, FullscreenStateConfig, UserConfig},
   windows::{
     commands::{
-      ignore_window, move_window_in_direction, move_window_to_workspace,
-      resize_window, set_window_size, update_window_state,
+      ignore_window, move_all_window_to_workspace,
+      move_window_in_direction, move_window_to_workspace, resize_window,
+      set_window_size, update_window_state,
     },
     traits::WindowGetters,
     WindowState,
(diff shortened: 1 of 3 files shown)


packages/wm/src/app_command.rs
```diff
@@ -188,6 +189,7 @@ pub enum InvokeCommand {
     #[clap(long)]
     direction: Direction,
   },
+  MoveAll(InvokeMoveCommand),
```
lars-berger, line 192: using the `InvokeMoveCommand` here will cause its flags to show as flags for `move-all` (e.g. `move-all --direction <direction>` even though that's not implemented)

would suggest creating a separate struct for `move-all`'s arguments

packages/wm/src/windows/commands/move_window_to_workspace.rs
```diff
@@ -13,6 +13,28 @@ use crate::{
...
 };
 
+pub fn move_all_window_to_workspace(
+  window: WindowContainer,
+  target: WorkspaceTarget,
+  state: &mut WmState,
+  config: &UserConfig,
+) -> anyhow::Result<()> {
+  let current_workspace = window.workspace().context("No workspace.")?;
+  let _ = current_workspace
```
lars-berger, line 23: should instead loop over the top-level children of the workspace (this will include both windows + split containers) and call `move_container_within_tree` on them to the target workspace.



pseudo-code:

```rust
let current_workspace = ...
let target_workspace = ...
let children = current_workspace.children();

for children {
  // move workspace children to target workspace
  move_container_within_tree( ...)

  // yoinked from L44-52 of `move_workspace_in_direction.rs`
  window.set_has_pending_dpi_adjustment(true);
  
  window.set_floating_placement(
    window
      .floating_placement()
      .translate_to_center(&workspace.to_rect()?),
  );
}
```

lars-berger (changes requested):

Issue #806, opened 2024-10-21
[Feature Request] A way to set an initial maximum window size.


EzequielBruni:
### Describe the problem/motivation

I have a wide monitor. As in 5120x1440p, 32:9. When I have any single window open, I really don't want to have it cover my full screen. Is there any way to set an initial maximum window size?

### Describe the solution you'd like

A simple setting in the `config.yaml` file to have an initial maximum width and height for new windows, which can then be resized manually as needed. These initial maximum dimensions should only be applied when there are no other windows open, or as long as there is empty space left on the screen. Widths and heights could be set by pixel dimensions or percentages.

### Alternatives considered

I could use per-app rules in AltSnap to enforce the max-width on a per-app basis, I suppose, but I don't actually know how it would work with GlazeWM and AltSnap essentially fighting over my window sizes.

DreamMaoMao:
I think the screen state should occupy the entire window, according to the wm rules under linux, it is generally allowed to set the window rules to specify the initial size of the floating window

EzequielBruni:
> I think the tiling state should occupy the entire window, according to the wm rules under linux, it is generally allowed to set the window rules to specify the initial size of the floating window

That's all well and good until you open up an app or page that doesn't have a dark mode for some reason and get flashbanged by your own monitor. And this isn't Linux. And isn't Linux supposed to be about choice, in any case? I'm not asking for this as a philosopher, or expert on wm rules. I just want the choice of not having every first window go automatically full screen on a monitor where that actually hurts the experience.

If y'all can't make that happen, just decide not to, or it's not a priority, that's fair. Y'all are making this for free, and I make no demands. But it seems silly to let an arbitrary rule like that get in the way of better UX, even if my situation is admittedly an edge case.

DreamMaoMao:
I don't think you understand the meaning of tiling, and this is not mandatory, you said that the description of things are so-called floating mode, you can not use tiling mode, and then set all the Windows to the initial state of floating, and then preset the size of the floating window

DreamMaoMao:
You have to use both tiling mode and a preset window size, which can be quite confusing

ShortArrow:
@DreamMaoMao @EzequielBruni It seems you're confusing fullscreen as a tiling window with fullscreen as a floating window.

PR #817, opened 2024-10-30
feat: add new `position` command; allow size and position to be set with `set-floating` command

DreamMaoMao:
This pr allows you to set size,position and centered status of the window when setting it to float, so that the launch size,position and center can be specified in the window open rule

### DIFF
--- a/packages/wm/src/app_command.rs
+++ b/packages/wm/src/app_command.rs
@@ -443,19 +443,14 @@ impl InvokeCommand {
       }
       InvokeCommand::Position(args) => {
         match subject_container.as_window_container() {
-          Ok(window) => {
-            match args.centered {
-              true => set_window_position_to_center(window, state),
-              false => {
-                set_window_position(
-                  window,
-                  args.x_pos.clone(),
-                  args.y_pos.clone(),
-                  state,
-                )
-              }
-            }
-
+          Ok(window) => match args.centered {
+            true => set_window_position_to_center(window, state),
+            false => set_window_position(
+              window,
+              args.x_pos.clone(),
+              args.y_pos.clone(),
+              state,
+            ),
           },
           _ => Ok(()),
         }
@@ -482,8 +477,7 @@ impl InvokeCommand {
         Ok(window) => {
           let floating_defaults =
             &config.value.window_behavior.state_defaults.floating;
-          let is_centered = 
-            centered.unwrap_or(floating_defaults.centered);
+          let is_centered = centered.unwrap_or(floating_defaults.centered);
 
           let window = update_window_state(
             window.clone(),
@@ -509,11 +503,11 @@ impl InvokeCommand {
             set_window_position_to_center(window, state)?;
           } else if x_pos.is_some() || y_pos.is_some() {
             set_window_position(
-              window.clone(), 
+              window.clone(),
               x_pos.clone(),
-              y_pos.clone(), 
+              y_pos.clone(),
               state,
-            )?;                
+            )?;
           }
 
           Ok(())
--- a/packages/wm/src/app_command.rs
+++ b/packages/wm/src/app_command.rs
@@ -3,6 +3,7 @@ use std::{iter, path::PathBuf};
 use anyhow::{bail, Context};
 use clap::{error::KindFormatter, Args, Parser, ValueEnum};
 use serde::{Deserialize, Deserializer, Serialize};
+use tokio::net::windows;
 use tracing::{warn, Level};
 use uuid::Uuid;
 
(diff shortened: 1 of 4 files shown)


DreamMaoMao:
maybe close:https://github.com/glzr-io/glazewm/issues/806

DreamMaoMao:
```yaml
  - commands: ["set-floating --centered --width=1500 --height=900"]
    match:
      - window_process: { regex: 'cloudmusic' }
```

https://github.com/user-attachments/assets/aa834e7e-0e61-46e9-9e44-3095752e5118

packages/wm/src/app_command.rs
```diff
@@ -441,11 +453,45 @@ impl InvokeCommand {
...
+          }
+
+          if is_centered {
+            let workspace =
+              window.workspace().context("no workspace find.")?;
+            window.set_floating_placement(
+              floating_placement
+                .translate_to_center(&workspace.to_rect()?),
+            )
+          } else if xpos.is_some() || ypos.is_some() || width.is_some() || height.is_some() {
```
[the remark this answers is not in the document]
DreamMaoMao: I think the one or two or three things you said, isn't that my current logic? 
I am new to rust, but I have tested these parameters one by one, and it is consistent with the logic you said above.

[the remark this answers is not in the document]
DreamMaoMao: In addition, set_window_size only sets the size and does not set the position. If the command of setting another position is taken, it will change the drawing action once into two draws, that is, change the size first, and then change the position. Is it inappropriate from the perspective of efficiency?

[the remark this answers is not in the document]
DreamMaoMao: done

packages/wm/src/app_command.rs
```diff
@@ -195,6 +195,18 @@ pub enum InvokeCommand {
 
     #[clap(long, default_missing_value = "true", require_equals = true, num_args = 0..=1)]
     centered: Option<bool>,
+
+    #[clap(long, require_equals = true)]
+    xpos: Option<i32>,
```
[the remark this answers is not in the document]
DreamMaoMao: done

packages/wm/src/windows/commands/set_window_position.rs
```diff
@@ -0,0 +1,87 @@
...
+        set_floating_window_position(
+          window,
+          centered,
+          target_x_pos,
+          target_y_pos,
+          state,
+        )?;
+      }
+    }
+  }
```
lars-berger, line 37: can remove the `match` statement and just have the `if matches! ...`

packages/wm/src/app_command.rs
```diff
@@ -438,25 +451,58 @@ impl InvokeCommand {
...
                 .unwrap_or(floating_defaults.shown_on_top),
             }),
             state,
             config,
           )?;
 
+          if width.is_some() || height.is_some() {
+            set_window_size(window.clone(),
+            width.clone(),
+            height.clone(),
```
lars-berger, line 494: no need to clone width/height

packages/wm/src/app_command.rs
```diff
@@ -189,12 +189,25 @@ pub enum InvokeCommand {
     direction: Direction,
   },
   Resize(InvokeResizeCommand),
+  Position(InvokePositionCommand),
```
lars-berger, line 192: move `Position(InvokePositionCommand),` below `MoveWorkspace ...` so it's alphabetical

packages/wm/src/app_command.rs
```diff
@@ -801,3 +847,16 @@ pub struct InvokeResizeCommand {
...
+#[group(required = true, multiple = true)]
+pub struct InvokePositionCommand {
+  #[clap(long, action)]
+  centered: bool,
+
+  #[clap(long, allow_hyphen_values = true)]
+  x_pos: Option<LengthValue>,
+
+  #[clap(long, allow_hyphen_values = true)]
+  y_pos: Option<LengthValue>,
```
lars-berger, line 861: change both to `Option<i32>` where it's a positive/negative coordinate

packages/wm/src/windows/commands/set_window_position.rs
```diff
@@ -0,0 +1,87 @@
...
+  },
+  windows::{
+    traits::WindowGetters, NonTilingWindow, WindowState,
+  },
+  wm_state::WmState,
+};
+
+pub fn set_window_position(
+  window: WindowContainer,
+  centered: bool,
```
lars-berger, line 17: remove `centered: bool` arg (only allowing target x/y pos) and instead do `window_rect.translate_to_center(&workspace.to_rect()?)` outside this fn

packages/wm/src/app_command.rs
DreamMaoMao: It doesn't seem to be possible to pass a reference value to option

lars-berger (approved):

PR #818, opened 2024-10-30
feat(config): Window Animations by config

veryard:
Added the ability to turn off window animations by default via the config, by default it is enabled like it was before.

Ps. Please let me know what formatter to use on this, I use VSCode default one and it appears to have changed something, sorry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,9 @@ general:
   # workspace when focusing the current workspace.
   toggle_workspace_on_refocus: false
 
+  # Whether to animate windows when switching between focused workspaces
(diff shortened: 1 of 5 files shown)


lars-berger:
Just adding this comment to document the solution from Discord:

Turns out [SystemParametersInfoW](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-systemparametersinfow) can take a flag to persist the changes between sessions (via `SPIF_UPDATEINIFILE` flag). Should then be able to avoid having to toggle this on every restart

PR #819, opened 2024-10-31
feat: add `--hide-window` flag to allow `shell-exec` to run commands in background

DreamMaoMao:
This will allow some terminal commands to be executed without pop-ups, such as powershell,nushell

before：

https://github.com/user-attachments/assets/043783e5-7dfb-4117-973e-5052cfb623d0

apply this pr:
```powershell
  - commands: ["shell-exec --hide-window nu ~/script/clash.nu"]
    bindings: ["Ctrl+Enter"]
```

https://github.com/user-attachments/assets/3c2000fd-e02e-41e0-baad-96985add4e78

### DIFF
--- a/packages/wm/src/app_command.rs
+++ b/packages/wm/src/app_command.rs
@@ -212,7 +212,7 @@ pub enum InvokeCommand {
   ShellExec {
     #[clap(long, action)]
     hide_window: bool,
-    
+
     #[clap(required = true, trailing_var_arg = true)]
     command: Vec<String>,
   },
--- a/packages/wm/src/app_command.rs
+++ b/packages/wm/src/app_command.rs
@@ -210,9 +210,9 @@ pub enum InvokeCommand {
     visibility: TitleBarVisibility,
   },
   ShellExec {
-    #[clap(long, default_missing_value = "true", require_equals = true, num_args = 0..=1)]
-    hide_window: Option<bool>,
-
+    #[clap(long, action)]
+    hide_window: bool,
+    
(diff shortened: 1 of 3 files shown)


packages/wm/src/app_command.rs
```diff
@@ -530,8 +533,15 @@ impl InvokeCommand {
           _ => Ok(()),
         }
       }
-      InvokeCommand::ShellExec { command } => {
-        shell_exec(&command.join(" "))
+      InvokeCommand::ShellExec { 
+        hide_window,
+        command 
+      } => {
+        if let Some(true) = hide_window {
```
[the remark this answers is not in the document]
DreamMaoMao: make sense

packages/wm/src/app_command.rs
```diff
@@ -210,6 +210,9 @@ pub enum InvokeCommand {
     visibility: TitleBarVisibility,
   },
   ShellExec {
+    #[clap(long, default_missing_value = "true", require_equals = true, num_args = 0..=1)]
+    hide_window: Option<bool>,
```
lars-berger, line 214: can be simplified by removing the `Option`: (default will be `false` if flag isn't specified)
```rust
   #[clap(long, action)]
   hide_window: bool,
```

lars-berger:
Great idea 👌

Only a minor code style fix, otherwise looks good!

packages/wm/src/app_command.rs
DreamMaoMao: optimize done

PR #821, opened 2024-11-02
fix: set focus to valid desktop window in an empty workspace

DreamMaoMao:
When switching to an empty workspace, we should logically focus on the desktop window, but in practice it fails, leaving us in a state where no window is focused, resulting in some window-event-based applications working abnormally, such as ahk:

desktop windows focus fail:

the windows message catch tool message: "can't get any focus window"

https://github.com/user-attachments/assets/9595f96c-398f-4c26-bfcd-e9fd292e87da

apply this pr:

https://github.com/user-attachments/assets/dd426f36-6e73-420d-9803-3f9304c78dd6

### DIFF
--- a/packages/wm/src/common/platform/platform.rs
+++ b/packages/wm/src/common/platform/platform.rs
@@ -19,9 +19,9 @@ use windows::{
       },
       WindowsAndMessaging::{
         CreateWindowExW, DispatchMessageW, GetAncestor, GetCursorPos,
-        FindWindowA, GetForegroundWindow, GetMessageW, MessageBoxW,
+        GetDesktopWindow, GetForegroundWindow, GetMessageW, MessageBoxW,
         PeekMessageW, PostThreadMessageW, RegisterClassW, SetCursorPos,
-        SystemParametersInfoW, TranslateMessage, WindowFromPoint,
+        SystemParametersInfoW, TranslateMessage, WindowFromPoint, GetShellWindow,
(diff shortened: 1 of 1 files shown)


packages/wm/src/common/platform/platform.rs
```diff
@@ -51,7 +51,9 @@ impl Platform {
 
   /// Gets the `NativeWindow` instance of the desktop window.
   pub fn desktop_window() -> NativeWindow {
-    let handle = unsafe { GetDesktopWindow() };
+    let  handle = unsafe {
+      FindWindowA(PCSTR("Progman\0".as_ptr() as *const u8), PCSTR::null())
+    };
```
lars-berger, line 56: think this can be simplified with:
```rust
    let handle = unsafe { GetShellWindow() };
```
should also handle the edge case of users who don't run explorer.exe:
```rust
    // Get Explorer wallpaper window (i.e. "Progman"). Default to the
    // desktop window on which the wallpaper window sits above for edge
    // case where Explorer isn't running.
    let handle = unsafe { GetShellWindow() }
      .unwrap_or_else(|| unsafe { GetDesktopWindow() });

Comment (DreamMaoMao) [reply packages/wm/src/common/platform/platform.rs]:
```rust
  pub fn desktop_window() -> NativeWindow {
    let handle = unsafe { GetDesktopWindow() };
    let mut handle = unsafe { GetShellWindow() };
    if handle == HWND(0) {
      handle = unsafe { GetDesktopWindow() };
    }
```

the return of GetShellWindow()  can't use unwrap_or_else, so i change to this.

Issue #790, opened 2024-10-13
[Bug] Regression in GlazeWM 3.4.0 overlaying with Taskbar


anthonymendez:
### Describe the bug

Hello, I realize that this may not be the right place but I figured I report this considering GlazeWM users may also use Windhawk or other Taskbar modifying programs :)

With the recent update, I noticed that the overlay isn't working as expected with my custom taskbar. I've modified the taskbar with Windhawk using:
* Taskbar Clock Customization
* Taskbar height and icon size
* Taskbar Labels for Windows 11
* Taskbar on Top for Windows 11
* Windows 11 Taskbar Styler 

The big issue is that the windows go under the taskbar on my main monitor. Then when I adjust the padding to compensate for this, my 2nd monitor with no taskbar has a large gap at the top with the programs.

Version 3.3.0 Example:
![{601BEB82-FE81-4FDF-B14C-E8DEA705C965}](https://github.com/user-attachments/assets/9bcf0eb7-0fde-44f5-91fd-842244b5e4b6)

Version 3.4.0 Example:
![{DF409530-1560-43EB-98E4-D8D56C1F67BC}](https://github.com/user-attachments/assets/e56f6450-eadb-4084-8f04-76f0ea2c77ff)

My config.yaml:
[config.yaml.txt](https://github.com/user-attachments/files/17351869/config.yaml.txt)

### Version number

3.4.0

Geobert:
I can confirm with StartAllBack on Win 11

isaac-varg:
Also having this issue with  Windows 11 and Start11 with the taskbar positioned on top.
![image](https://github.com/user-attachments/assets/fe5d2c48-438d-4205-b713-46868c2b4c20)

JonasWischeropp:
Does it help to set `gaps.scale_with_dpi` to `false`?
```yaml
gaps:
    scale_with_dpi: false
```

isaac-varg:
Configuring `scale_with_dpi` to `false` did not seem to help. 

I attempted to change the `gaps.outer_gap.top` to a higher value to avoid the taskbar and the issue persisted. The issue did resolve when I move my taskbar to be positioned at the bottom.

Dovias:
Can confirm. As far i've tested it it seems to happen only in those cases where specified top outer gap is being set at around 50 px. Strangely enough this matches the windows 11 taskbar size for me. Top taskbar alignment are wonky on Windows 11 sadly :(

[...]

Issue #640, opened 2024-08-05
[Bug] Changing from maximized -> fullscreen via `alt+f` causes flickering


AmineDjeghri:
When you maximize a window (using the window button next to the close button) then press ``ALT+F`` , the windows start shaking then crash.

monoelh:
With me it's not shaking, but the margins are screwed (and the app is not crashing).
`max with button -> alt+f (now top margin bad) -> alt+f (now fullscreen again) -> alt+f (now margins are fine)`

Interestingly, this is working correctly for windows explorer.

That being said, other hotkeys are not working for me (e.g. `alt+shift+a/s/d/f`).

AmineDjeghri:
steps to reproduce : 

- open an app (for example bloc-notes)
- click on 'CTRL+F' to make the window in full screen (the window is in full screen but not maximized)
- click on the maximize button of the window
- click on the unmaximize button (the same buttton)


Temporary fix : 
Change maximized to true 

![image](https://github.com/user-attachments/assets/9c61190a-034a-4bb0-95bd-6151f98a4583)

AmineDjeghri:
![Notepad_Prny8GiVi1](https://github.com/user-attachments/assets/1e80ddb6-8efd-4cc7-9e69-8efbec6aaa1a)

PR #823, opened 2024-11-03
fix: regression in layouting with fullscreen state and various taskbar alignments

Dovias:
# This PR implements fixes for #790 and #640 issues

I've tested **StartAllBack** and **Windhawk's top taskbar mod** by @m417z , but I am pretty confident **ExplorerPatcher** would also work due to the nature of **StartAllBack**'s use of classic taskbar.

**Note:** I've learned Rust basics just because how annoying this bug was for me, so if you see some spots which look kind of bad in the code, please review it 🤣

**Implementation note**: This PR completely removes `Rect`'s `apply_delta()` and `apply_inverse_delta()` functions as they were suited only for `Window` calculations, completely ignoring `Workspace` delta requirements. All of the calculations required to fix this in this PR were made in-place by constructing a `Rect` via`Rect::from_ltrb()` function instead.

# Before (this issue persists since at least GlazeWM 3.4.0):

https://github.com/user-attachments/assets/49232096-19f2-467a-a81d-8d88800ff994

https://github.com/user-attachments/assets/12377265-a8ef-4295-8463-ee648e9ff092

# After:

https://github.com/user-attachments/assets/7c464d72-37ea-48bb-b2e3-ce8c375e8b33

https://github.com/user-attachments/assets/b71bb32e-8712-44e0-bc70-8ef2a52aad97

### DIFF
--- a/packages/wm/src/common/commands/platform_sync.rs
+++ b/packages/wm/src/common/commands/platform_sync.rs
@@ -5,7 +5,7 @@ use tokio::task;
 use tracing::warn;
 
 use crate::{
-  common::{platform::Platform, DisplayState},
+  common::{platform::Platform, DisplayState, Rect},
   containers::{
     traits::{CommonGetters, PositionGetters},
     Container, WindowContainer,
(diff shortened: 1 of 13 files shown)


BlenderEngineer:
Very cool. Good job

Dovias:
Added a fix for fullscreen overflow under taskbar (this issue persists even with a bottom taskbar on GlazeWM 5.0)

Dovias:
Added a fix for fullscreen sync and automatic minimization for fullscreen windows, if the non-floating window is being used

Dovias:
After discussing with core team, we've decided that this PR is too big to be merged. This PR will be separated into three separate PR's which represent according fixes.

PR #828, opened 2024-11-03
feat: add `wm-toggle-pause` command

### LINKED ISSUE #825: [Feature Request] New mode for really pausing glazewm

The example config comes with a pause mode, which essentially disables all keybindings but one.

While I don't see much use for this feature, I real pausing mode (we could call it `halt`) would be nice, which makes glazewm temporarily stop managing windows.

Why this is useful?
- When playing games (especially in a lower resolution that desktop) and having multiple monitors, glazewm moves windows for some time and it's difficult to get back to the game
- Sometimes when moving windows, glazewm randomly acts out and the whole screen is flickering
- Also, when you want to let others temporarily use your PC and they don't know glaze.

While the former two are bugs and will probably get fixed sometime, this would be a neat workaround.

A new command which pauses glazewm:
`toggle-pause` (or, to not confuse with the binding mode pause from the examples, `toggle-halt`)

It would be nice if zebar could query this information and display an icon or grey out the bar.

### Alternatives considered

Making a `shell-exec` command to kill the glazewm process. However, you cannot start it again by a key combination. It also takes some time to start and it will try to launch a second zebar process as well.

michidk:
This is my try to implement the pause command I proposed in https://github.com/glzr-io/glazewm/issues/825.

Basically, this adds the command `wm-toggle-pause` which pauses glazewm from managing windows.

Feedback is very welcome. This is my first contribution to glazewm and I'm not too familiar with the codebase yet.


lars-berger:
Ay this is super cool 👌 Thanks for putting together the IPC side of it as well.

A downside with returning early in all these platform events is that state will be out of sync on resume. Like say a monitor was unplugged, the WM would still think it has all its monitors.

Perhaps it'd be better to instead skip redraws if the WM is paused:
```rust
// wm.rs - https://github.com/glzr-io/glazewm/blob/940bb4cc474898aa3b1ffda2f29720e37bb7a16c/packages/wm/src/wm.rs#L128
..
    if state.is_paused {
      platform_sync(state, config)?;
    }
```
and then also disable keyboard/mouse listener on pause:
```rust
// main.rs - https://github.com/glzr-io/glazewm/blob/940bb4cc474898aa3b1ffda2f29720e37bb7a16c/packages/wm/src/main.rs#L162
..
        if matches!(
          wm_event,
          WmEvent::UserConfigChanged { .. }
            | WmEvent::BindingModesChanged { .. }
            | WmEvent::PauseChanged { .. }
        ) {
          event_listener.update(
            &config,
            &wm.state.binding_modes,
            is_paused
          );
        }
```

michidk:
Hi @lars-berger,

My new implementation to block mouse events goes probably into the right direction.

But there is one problem with blocking key events: The keybinding to unpause won't also be recognized.
However, I went forward and implemented something.

If you think this is an alright approach, I would go and implement the last missing part:
https://github.com/michidk/glazewm/blob/pause/packages/wm/src/common/platform/keyboard_hook.rs#L365
(detecting if the keybind is for unpausing and then not blocking it).

michidk:
Done.

glazewm-js: https://github.com/glzr-io/glazewm-js/pull/15
zebar: https://github.com/glzr-io/zebar/pull/167

lars-berger:
All merged now! Thanks for getting it all nicely integrated with the NPM library + zebar as well 🙌 

Found a couple minor things while testing it out more thoroughly #870. Also did a super nitpicky rename to `is_paused` since that seems more consistent with our other boolean properties like `has_focus`, `is_displayed`...

Thanks again for the massive help on this ❤️

michidk:
I actually had used `is_paused` in some iterations as well, but then removed it again. 

I think right now we sometimes have still `paused` in the NPM library but I guess that's fine.

Thank you for finishing this up and sorry for missing the resize & IPC events!

PR #834, opened 2024-11-05
refactor: clippy fixes and working doctests

Miitto:
# Linting
Fixes most lint errors provided by clippy

Unused  variables / functions have not been fixed

Clippy detects `needless_return` on both `main` functions, which I believe is caused by `#[tokio::main]`, so these have been supressed.

# Doctests
The following are no longer run:
- `wm\src\containers\commands\flatten_child_split_containers.rs:11` : `flatten_child_split_containers` has been ignored
- `wm\src\common\platform\platform.rs:340`: `parse_command` has been set to `no_run` as it requires files to exist on the test machine.

The following have been fixed so that they can be compiled and ran:
- `wm\src\common\tiling_direction.rs:19`: `inverse`
- `wm\src\common\tiling_direction.rs:35`: `from_direction`
- `wm\src\common\tiling_direction.rs:54`: `from_str`
- `wm\src\common\direction.rs:19` : `inverse`
- `wm\src\common\direction.rs:40` : `from_str`
- `wm\src\common\length_value.rs:52`: `from_str`

### DIFF
--- a/packages/watcher/src/main.rs
+++ b/packages/watcher/src/main.rs
@@ -12,6 +12,8 @@ use wm::{
 };
 
 #[tokio::main]
+// Clippy picks up on a return that is likely added by `#[tokio::main]`.
+#[allow(clippy::needless_return)]
 async fn main() -> anyhow::Result<()> {
   tracing_subscriber::fmt().init();
 
@@ -35,7 +37,7 @@ async fn main() -> anyhow::Result<()> {
 
       let managed_windows = managed_handles
         .into_iter()
-        .map(|handle| NativeWindow::new(handle))
+        .map(NativeWindow::new)
         .collect::<Vec<_>>();
 
       run_cleanup(managed_windows);
@@ -56,7 +58,7 @@ async fn query_initial_windows(
     .context("Failed to send window query command.")?;
 
   client
-    .client_response(&query_message)
+    .client_response(query_message)
     .await
     .and_then(|response| match response.data {
       Some(ClientResponseData::Windows(data)) => Some(data),
@@ -83,12 +85,12 @@ async fn watch_managed_handles(
     "sub -e window_managed window_unmanaged application_exiting";
 
   client
-    .send(&subscription_message)
+    .send(subscription_message)
     .await
     .context("Failed to send subscribe command to IPC server.")?;
 
   let subscription_id = client
-    .client_response(&subscription_message)
+    .client_response(subscription_message)
     .await
     .and_then(|response| match response.data {
       Some(ClientResponseData::EventSubscribe(data)) => {
--- a/packages/watcher/src/main.rs
+++ b/packages/watcher/src/main.rs
@@ -12,8 +12,6 @@ use wm::{
 };
 
 #[tokio::main]
-// Clippy picks up on a return that is likely added by `#[tokio::main]`.
-#[allow(clippy::needless_return)]
 async fn main() -> anyhow::Result<()> {
   tracing_subscriber::fmt().init();
 
--- a/packages/wm/src/common/platform/platform.rs
+++ b/packages/wm/src/common/platform/platform.rs
@@ -376,7 +376,7 @@ impl Platform {
     };
 
     let command_parts: Vec<&str> =
-      expanded_command.trim().split_whitespace().collect();
+      expanded_command.split_whitespace().collect();
 
     // If the command starts with double quotes, then the program name/path
     // is wrapped in double quotes (e.g. `"C:\path\to\app.exe" --flag`).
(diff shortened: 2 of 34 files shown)


packages/wm/src/workspaces/commands/focus_workspace.rs
```diff
@@ -11,7 +11,9 @@ use crate::{
   },
 };
 
-/// Focuses a workspace by a given target. This target can be a workspace
+/// Focuses a workspace by a given target.
+///
+/// This target can be a workspace
```
lars-berger, line 16: Comment gets cut off

packages/wm/src/user_config.rs
```diff
@@ -659,17 +661,19 @@ fn default_window_rule_on() -> Vec<WindowRuleEvent> {
   vec![WindowRuleEvent::Manage, WindowRuleEvent::TitleChange]
 }
 
-impl Default for CornerStyle {
-  fn default() -> Self {
-    CornerStyle::Default
-  }
-}
+// Both are now derrived, left in case it is needed in the future
```
lars-berger, line 664: fine with deleting these 👍

packages/wm/src/common/platform/platform.rs
```diff
@@ -337,16 +337,23 @@ impl Platform {
...
   ///
-  /// ```
-  /// let (prog, args) = parse_command("code .")?;
-  /// assert_eq!(prog, "code");
-  /// assert_eq!(args, ".");
+  /// ```no_run
+  /// # // Since this checks if the path exists, not sure how to test this.
+  /// # // Don't think you can `use` a struct method.
+  /// # use wm::common::platform::Platform;
+  /// fn main() -> anyhow::Result<()> {
```
lars-berger, line 344: would it work to have this inlined without the `main` fn?

should also be fine to remove the clarifying the comment here 👍

packages/watcher/src/main.rs
```diff
@@ -12,6 +12,8 @@ use wm::{
 };
 
 #[tokio::main]
+// Clippy picks up on a return that is likely added by `#[tokio::main]`.
+#[allow(clippy::needless_return)]
```
lars-berger, line 16: looks like this is fixed in clippy 0.1.80 https://github.com/rust-lang/rust-analyzer/issues/16542

packages/wm/src/containers/commands/flatten_child_split_containers.rs
```diff
@@ -8,7 +8,7 @@ use crate::containers::{
 /// parent container.
 ///
 /// For example:
-/// ```
+/// ```no_run,compile_fail,ignore
```
lars-berger, line 11: would it work to only use `ignore` here?

packages/wm/src/main.rs
```diff
@@ -46,28 +46,30 @@ mod workspaces;
 /// Conditionally starts the WM or runs a CLI command based on the given
 /// subcommand.
 #[tokio::main]
+// Clippy picks up on a return that is likely added by `#[tokio::main]`.
+#[allow(clippy::needless_return)]
 async fn main() -> Result<()> {
   let args = std::env::args().collect::<Vec<_>>();
   let app_command = AppCommand::parse_with_default(&args);
 
-  match app_command {
```
lars-berger, line 53: imo it's more semantic to `match` on the command here even though there are just 2 possible actions currently

lars-berger:
This is super nice!! Been meaning to add clippy linting for ages 🙌 

Would you happen to know what the recommended command would be to to add linting to CI?

Miitto:
Not sure how to add it to CI, but it can be manually run via `cargo clippy`.

Can be added to the tool chain via `rustup component add clippy` if it is not already present.


[Docs here](https://doc.rust-lang.org/stable/clippy/continuous_integration/index.html)


```
on: push
name: Clippy check

# Make sure CI fails on all warnings, including Clippy lints
env:
  RUSTFLAGS: "-Dwarnings"

jobs:
  clippy_check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - name: Run Clippy
        run: cargo clippy --all-targets --all-features

```
Taken from [here](https://doc.rust-lang.org/stable/clippy/continuous_integration/github_actions.html)

packages/wm/src/common/platform/platform.rs
Miitto: You can't use the ? for results without the main function. The code for the main function could likely be hidden in the docs by using the `# ` at the start of the line, then un-indenting the example.

lars-berger: ah I see - wouldn't mind if we just change these to `.unwrap()` instead then

packages/wm/src/containers/commands/flatten_child_split_containers.rs
Miitto: Yes, though it may? be worth it to keep `compile_fail` in addition, incase `cargo test --doc -- --ignored` is run in future. Though that is admittedly a rather niche case.

No run was entirely redundant.

lars-berger:
Merged 👍 

Will look into getting rid of those unused var warnings - would be super nice to enforce these lints for PR's. Ty for the info on adding it to CI!

PR #838, opened 2024-11-07
fix: regression in layouting with fullscreen state

Dovias:
# Fix regression in layouting with fullscreen state
This PR fixes an issue where window's fullscreen state would overflow under existing elements in the desktop. Currently GlazeWM doesn't take in-account the delta offsets of the working area of the monitor when calculating fullscreen state window size, thus graphical issues occur in the latest version of GlazeWM. This PR should fix it.

**Note**: This PR works best with #836 PR, since it also fixes issues with graphical issues, related to the delta offsets of working area.

# Before:

https://github.com/user-attachments/assets/f8c598a5-0f3d-4760-becc-b49668ec4ed0

# After:

https://github.com/user-attachments/assets/c81e7765-f880-4d6b-b698-a4a4c85bd652

### DIFF
--- a/packages/wm/src/common/events/handle_window_location_changed.rs
+++ b/packages/wm/src/common/events/handle_window_location_changed.rs
@@ -5,7 +5,7 @@ use crate::{
   common::{platform::NativeWindow, Rect},
   containers::{
     commands::{flatten_split_container, move_container_within_tree},
-    traits::{CommonGetters, PositionGetters},
+    traits::CommonGetters,
     WindowContainer,
   },
   try_warn,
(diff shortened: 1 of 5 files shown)


lars-berger:
Have been doing some testing, and I believe this issue is resolved now with #851 (while still preserving the functionality of fullscreen'ing to the full monitor size).

There's still good ideas here - as discussed on Discord, it'd be a nice addition to draw inspo from this PR and have some way of conditionally fullscreen'ing to either the full monitor size or the working area (e.g. via a new flag to `set-fullscreen`). Think it'd best to potentially tackle that in a separate PR though.

PR #842, opened 2024-11-09
feat: improve `--centered` flag behavior for `set-floating` and `toggle-floating`

DreamMaoMao:
```yaml
  - commands: ['toggle-floating --centered']
    bindings: ['alt+\']
```
if use `--centered` option in toggle-flating. it will be set position of the flaoting window to centered at first time.

 

https://github.com/user-attachments/assets/c0b29a3a-12e4-4b3d-918e-deef8955dfe3


lars-berger:
Hey @DreamMaoMao - I hope you don't mind but did a couple of tweaks: 1) check `has_custom_floating_placement` for `set-floating` so that it's consistent with `toggle-floating` 2) remove `has_custom_floating_placement` from the DTO (this gets emitted via our public IPC interface, and I think it's probs best if we just use it internally).

Thanks for the huge help on this🙏 Will go ahead and merge!

PR #865, opened 2024-11-27
feat: revert to keybindings for prev/next active workspace in sample config

DreamMaoMao:
Notice many people asking how this key is bound, and it is a common feature, so suggest adding them to the default sample configuration

### DIFF
--- a/resources/assets/sample-config.yaml
+++ b/resources/assets/sample-config.yaml
@@ -250,6 +250,12 @@ keybindings:
   - commands: ['focus --prev-workspace']
     bindings: ['alt+a']
 
+  # Focus the next/previous active workspace defined in `workspaces` config.
+  - commands: ['focus --next-active-workspace']
+    bindings: ['alt+e']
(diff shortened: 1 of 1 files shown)


lars-berger:
Ah I overlooked that the sample config still uses `--next-workspace` and `--prev-workspace` after PR #783. Even though both commands are useful, switching between active workspaces is likely the more common use case.

Since they're such similar commands, I don't think it's worth adding both to the sample config and I'd suggest we instead change `alt+a` and `alt+s` to use `--prev-active-workspace` and `--next-active-workspace` instead.

DreamMaoMao:
> Ah I overlooked that the sample config still uses `--next-workspace` and `--prev-workspace` after PR #783. Even though both commands are useful, switching between active workspaces is likely the more common use case.
> 
> Since they're such similar commands, I don't think it's worth adding both to the sample config and I'd suggest we instead change `alt+a` and `alt+s` to use `--prev-active-workspace` and `--next-active-workspace` instead.

makes sense. It's been replaced.

PR #889, opened 2024-12-16
fix: pause IPC event not emitting

michidk:
Fixes the pause IPC not working:

```sh
PS Microsoft.PowerShell.Core\FileSystem::\\wsl.localhost\Arch\home\michi\dev\glazewm\target\x86_64-pc-windows-gnu\debug> ./glazewm sub -e pause_changed
{"data":{"eventType":"pause_changed","isPaused":false},"error":null,"subscriptionId":"bf8e2089-487a-4cc9-8846-f5c9e7b17b82","success":true}
{"data":{"eventType":"pause_changed","isPaused":false},"error":null,"subscriptionId":"bf8e2089-487a-4cc9-8846-f5c9e7b17b82","success":true}
{"data":{"eventType":"pause_changed","isPaused":false},"error":null,"subscriptionId":"bf8e2089-487a-4cc9-8846-f5c9e7b17b82","success":true}
```

### DIFF
--- a/packages/wm/src/wm_state.rs
+++ b/packages/wm/src/wm_state.rs
@@ -435,7 +435,10 @@ impl WmState {
   /// from being emitted via IPC server before the initial state is
   /// prepared.
   pub fn emit_event(&self, event: WmEvent) {
-    if self.has_initialized && !self.is_paused {
+    if self.has_initialized
+      && (!self.is_paused
+        || matches!(event, WmEvent::PauseChanged { is_paused: _ }))
+    {
       if let Err(err) = self.event_tx.send(event) {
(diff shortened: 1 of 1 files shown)


michidk:
Im not too familiar with how keybindings work, but we don't unregister them in pause mode. So this fix won't solve that.
I think the best way to do that would be a shortcut which kills glazewm and starts it again instead of pausing.

I totally see you usecase. But it assumes that the other application tries to register it's hotkeys during app start (so it would need to be started when glazewm is paused), but when unpausing glazewm probably will not be able to register the hotkey again (unless all other apps which might register them are closed).

PR #921, opened 2025-01-17
ci: updated to newer winget release workflow

vikingnope:
Updated the workflow to a new commit version and changed the username to match the username that is currently being used. Should reduce potential errors in the future.


lars-berger:
Thank you 🙏

PR #937, opened 2025-01-22
fix: Window effects don't update on focus switch

Miitto:
Fix for window effects not updating when window focus changes.

At HEAD window effects such as border colors are not updating when window focus changes.

Issue seems? to be that the `handle_window_focused` is called after the focused container of `state` is updated to the new window.


lars-berger:
Dang, you caught that super fast. My b for letting this reach `main`, personally don't run window effects so it took me a while to catch

Just merged in the last PR I've got planned for the next v3.8 release btw, so it won't be long before a combined release of glazewm + zebar. LMK if you need any help from me aside from taking a look at your last couple of PR's 👌

PR #941, opened 2025-01-25
fix: float windows by default if they have `WS_POPUP` style

ChichiMsdk:
This is to set windows **_meant_** to have a specific size (but they are still resizable) to be in floating mode instead of tiling by default.

This could _maybe_ replace the workarounds using the window class for some explorer's popups. Would have to check if those windows have the 'WS_POPUP' flag or not.

Without this, many windows are improperly resized and makes the tiling manager almost unusable, especially with softwares like Visual Studio or Ghidra.


packages/wm/src/commands/window/manage_window.rs
```diff
@@ -211,6 +211,12 @@ fn window_state_to_create(
     ));
   }
 
+  if native_window.is_popup() {
```
lars-berger, line 214: could combine this into the `!native_window.is_resizable` check above

lars-berger (changes requested):

lars-berger:
Good point! Think this would make a lot of sense to add as a default 👌 

Thank you for the PR

ChichiMsdk (commented):
I am not sure if this is the correct syntax for an "or" condition in rust as I've never programmed in it. Hoping it is correct

lars-berger (approved):

lars-berger:
Awesome 🙏 Made a quick linting fix (hope it's ok I pushed directly to the branch).

I held off including this in the current release to give some more time for testing with different apps. From what I've seen in some quick initial testing though, it's looking good!

PR #959, opened 2025-02-02
feat: change monitor focus with `general.focus_follows_cursor`

jackssrt:
This PR makes the general.focus_follows_cursor code focus the monitor that the cursor is on when there isn't a window to focus. This makes it so if you focus monitor 1, move your cursor to monitor 2 and open a program, the program will open on monitor 2 and not on monitor 1.

Before:

https://github.com/user-attachments/assets/bd260c32-8add-442c-b518-0e5a4ff14aa9

After:

https://github.com/user-attachments/assets/22ce7026-f1fb-43ee-8a92-85125c3127a3

This has been annoying me for a while, so I decided to fix it...

I don't know if it would've been better to make it focus the workspace under the cursor or not. Since there can only be one workspace displayed per monitor, I chose to make it focus the monitor.
Just let me know if you think it would be better to focus the workspace instead!

### DIFF
--- a/packages/wm-platform/src/platform.rs
+++ b/packages/wm-platform/src/platform.rs
@@ -160,6 +160,19 @@ impl Platform {
     Ok(NativeWindow::new(handle.0))
   }
 
+  pub fn monitor_from_point(
+    point: &Point,
+  ) -> anyhow::Result<NativeMonitor> {
+    let monitors = native_monitor::available_monitors()?;
+    monitors
+      .iter()
+      .find_map(|monitor| {
+        let rect = monitor.rect().ok()?;
+        rect.contains_point(point).then(|| monitor.clone())
+      })
+      .context("Monitor not found")
+  }
+
   /// Gets the mouse position in screen space.
   pub fn mouse_position() -> anyhow::Result<Point> {
     let mut point = POINT { x: 0, y: 0 };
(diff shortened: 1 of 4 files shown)


jackssrt:
This causes some problems with the ShareX screenshot overlay. When you open it and move your cursor to another monitor it steals focus away from the overlay. This happens even tough the overlay spans all your monitors, and it *should* be detected by the window focusing code...

Maybe the Windows API function for finding windows overlapping a point doesn't really work, and we should instead make our own implementation? Or maybe we should add a command to always keep a window focused?

lars-berger (changes requested):

jackssrt:
[80b01dc](https://github.com/glzr-io/glazewm/pull/959/commits/80b01dce60d779499b0e48cb0754b02378f4f3f6) has some problems right now when moving the mouse between monitors and windows. I think the is_focus_synced field isn't being set to true in all the cases it should be. I will look into it.

lars-berger:
> [80b01dc](https://github.com/glzr-io/glazewm/pull/959/commits/80b01dce60d779499b0e48cb0754b02378f4f3f6) has some problems right now when moving the mouse between monitors and windows. I think the is_focus_synced field isn't being set to true in all the cases it should be. I will look into it.

Put in a small fix commit for this just now - could you check if it's fully working on your end? Did some quick testing and it all seems to be behaving nicely now

lars-berger (approved):

jackssrt:
It seems to be working now, thanks for fixing! And sorry for taking so long to reply… You can merge it now.

PR #960, opened 2025-02-02
fix: fix truncated window title log panic

jackssrt:
Fixes a panic that only happens in development. If a window title contains an emoji or something right where it gets truncated by `(WindowContainer as std::fmt::Display)::fmt`, a panic will occur. 
Truncating a `&str` can cause it to contain invalid utf-8, which is not allowed in Rust strings. This PR adds some code to make sure it finds a safe character boundary before truncating.
Based on https://users.rust-lang.org/t/truncate-str-to-n-utf-8-chars/86626/14


jackssrt:
Oops... that was already fixed apparently

PR #962, opened 2025-02-02
chore: add vscode settings

jackssrt:
This PR adds a settings.json file in the .vscode folder. This will make all VS Code users automatically get the correct indentation settings for this project and use clippy as their rust-analyzer check command, as mentioned in https://github.com/glzr-io/glazewm/blob/main/CONTRIBUTING.md#tips .

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "rust-analyzer.check.command": "clippy",
+}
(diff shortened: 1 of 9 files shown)


.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rust-analyzer.check.command": "clippy",
+  "editor.insertSpaces": true,
+  "editor.tabSize": 2,
+}
```
lars-berger, line 5: we could also add:
```json
  "files.insertFinalNewline": true,
  "files.trimTrailingWhitespace": true,
```

lars-berger (changes requested):

lars-berger (approved):

lars-berger:
Thanks for the fixes🙏 Merged this into `main` just now!

PR #963, opened 2025-02-02
feat: add `focus --container-id <id>` command

jsfr:
This PR introduces a new command to focus a specific container by ID.

This allows to write scripts with the CLI to e.g. focus windows from a specific application.


packages/wm/src/commands/container/focus_container_id.rs
```diff
@@ -0,0 +1,20 @@
+use anyhow::Context;
+use uuid::Uuid;
+
+use super::set_focused_descendant;
+use crate::wm_state::WmState;
+
+pub fn focus_container_id(
```
lars-berger, line 7: nit: rename fn and filename to `focus_container_by_id`

lars-berger (changes requested):

lars-berger:
Thanks for the PR ❤️ Would def be useful to set specific focus, especially for IPC use cases.

There was a lint that just got added to `clippy` nightly which I've just put in a fix for (#967). If you sync with the `main` branch, the lint check flow should pass again 👍 

Other than that LGTM

packages/wm/src/commands/container/focus_container_id.rs
jsfr: Done ✅

jsfr:
@lars-berger I've addressed the comments :) Let me know if there's anything else missing.

lars-berger (approved):

lars-berger:
Perfect 🙌 thank you!

It's looking like the next release will be in 1.5-2 weeks. Doing a big Zebar feature and then I'll do a combined release of GlazeWM+Zebar with this PR included 👍

PR #980, opened 2025-02-21
feat: add `focus_workspace_on_current_monitor` and `swap_monitor` commands

Im-shamo:
Added two new commands

## `focus_workspace_on_current_monitor`
This is a slight addition to `focus_workspace` command

This focuses a workspace by a given target.
- If the target workspace and focused workspace is in the same monitor, it does the same thing as `focus_workspace` function.
- If the target workspace is on a different monitor, it will move target workspace to the focused monitor and focuses it.
- If the target workspace is displayed on a different monitor, it will swap the target workspace and the focused workspace. After that, focus the target workspace.

## `swap_monitor`
Swaps the active workspace on `monitor a` and `monitor b`.

- By default, focus stay the same. However, if `stay_on_monitor` is enable, the focus with stay on the same monitor.

### DIFF
--- a/packages/wm-common/src/app_command.rs
+++ b/packages/wm-common/src/app_command.rs
@@ -160,6 +160,7 @@ pub enum InvokeCommand {
   Close,
   Focus(InvokeFocusCommand),
   FocusWorkspaceOnCurrentMonitor(InvokeFocusWorkspaceOnCurrentMonitorCommand),
+  SwapWorkspace(InvokeSwapWorkspaceCommand),
   Ignore,
   Move(InvokeMoveCommand),
   MoveWorkspace {
@@ -351,6 +352,20 @@ pub struct InvokeFocusWorkspaceOnCurrentMonitorCommand {
 
   #[clap(long)]
   pub recent_workspace: bool,
+}
+
+#[derive(Args, Clone, Debug, PartialEq, Serialize)]
+#[group(required = true, multiple = true, requires_all=["monitor_1", "monitor_2"])]
+#[allow(clippy::struct_excessive_bools)]
+pub struct InvokeSwapWorkspaceCommand {
+  #[clap(long)]
+  pub monitor_1: Option<usize>,
+
+  #[clap(long)]
+  pub monitor_2: Option<usize>,
+
+  #[clap(long)]
+  pub stay_on_monitor: bool
 }
 
 #[derive(Args, Clone, Debug, PartialEq, Serialize)]
--- a/packages/wm-common/src/app_command.rs
+++ b/packages/wm-common/src/app_command.rs
@@ -159,6 +159,7 @@ pub enum InvokeCommand {
   AdjustBorders(InvokeAdjustBordersCommand),
   Close,
   Focus(InvokeFocusCommand),
+  FocusWorkspaceOnCurrentMonitor(InvokeFocusWorkspaceOnCurrentMonitorCommand),
   Ignore,
   Move(InvokeMoveCommand),
   MoveWorkspace {
@@ -312,6 +313,29 @@ pub struct InvokeFocusCommand {
 
   #[clap(long)]
   pub monitor: Option<usize>,
+
+  #[clap(long)]
+  pub next_active_workspace: bool,
+
+  #[clap(long)]
+  pub prev_active_workspace: bool,
+
+  #[clap(long)]
+  pub next_workspace: bool,
+
+  #[clap(long)]
+  pub prev_workspace: bool,
+
+  #[clap(long)]
+  pub recent_workspace: bool,
+}
+
+#[derive(Args, Clone, Debug, PartialEq, Serialize)]
+#[group(required = true, multiple = false)]
+#[allow(clippy::struct_excessive_bools)]
+pub struct InvokeFocusWorkspaceOnCurrentMonitorCommand {
+  #[clap(long)]
+  pub workspace: Option<String>,
 
   #[clap(long)]
   pub next_active_workspace: bool,
--- a/packages/wm/src/commands/workspace/focus_workspace_on_current_monitor.rs
+++ b/packages/wm/src/commands/workspace/focus_workspace_on_current_monitor.rs
@@ -159,6 +159,10 @@ fn move_and_focus(
     .queue_container_to_redraw(focused_workspace.clone())
     .queue_cursor_jump();
 
+  state.emit_event(WmEvent::WorkspaceUpdated {
+    updated_workspace: target_workspace.to_dto()?
+  });
+
   state.recent_workspace_name = Some(target_workspace.config().name);
 
   Ok(())
--- a/packages/wm/src/commands/workspace/focus_workspace_on_current_monitor.rs
+++ b/packages/wm/src/commands/workspace/focus_workspace_on_current_monitor.rs
@@ -1,0 +1,252 @@
+use anyhow::Context;
+use tracing::info;
+use wm_common::WmEvent;
+
+use super::activate_workspace;
+use crate::{
+  commands::{
+    container::{move_container_within_tree, set_focused_descendant},
+    workspace::{deactivate_workspace, sort_workspaces},
+  },
+  models::WorkspaceTarget,
+  traits::{CommonGetters, PositionGetters, WindowGetters},
+  user_config::UserConfig,
+  wm_state::WmState,
+};
+
+/// Focuses a workspace by a given target.
+///
+/// If the target workspace and focused workspace is in the same monitor,
+/// does the same thing as `focus_workspace` function.
+///
+/// If the target workspace is on a different monitor,
+/// it will move target workspace to the focused monitor and focuses it.
+///
+/// If the target workspace is displayed on a different monitor,
+/// it will swap the target workspace and the focused workspace and focuse
+/// the target workspace.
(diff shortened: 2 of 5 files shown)


Im-shamo:
Take a look to see if the way I am implementing it make sense.

packages/wm-common/src/app_command.rs
```diff
@@ -159,6 +159,8 @@ pub enum InvokeCommand {
   AdjustBorders(InvokeAdjustBordersCommand),
   Close,
   Focus(InvokeFocusCommand),
+  FocusWorkspaceOnCurrentMonitor(InvokeFocusWorkspaceOnCurrentMonitorCommand),
+  SwapWorkspace(InvokeSwapWorkspaceCommand),
```
lars-berger, line 163: For `SwapWorkspace`, since it currently takes in both the current and a target monito, it means that it can essentially only be used for programmatic use or if the user hand tailors their config. imo it'd be nicer if it simply took a `--monitor <monitor_index>` or `--direction <direction>` arg (or both) and then operated on the subject container's parent monitor. With IPC, the subject container can be set to whatever window/workspace/monitor the user wants to operate on, so you'd still have the full functionality of swapping whichever 2 monitors you'd like as in your current implementation

lars-berger (changes requested):

lars-berger:
Hey, thanks for the last couple PRs! These are super well written; you clearly know what you're doing when it comes to Rust 👌 

Both of the features in this PR are really nice. I think they can be structured in a way where it'd be more flexible for the future - left a couple PR review comments just now.

Side note: it's easier to review if features are separated out into different PR's (e.g. one for `swap-workspace` and one for `focus-workspace-on-current-monitor`). It's totally fine to leave it as is, just a heads up for future PRs to GlazeWM (or other open source projects) 👍

Im-shamo:
Much thanks for the input. I will work on it when I have the time! :)

packages/wm-common/src/app_command.rs
Im-shamo: Good suggestion. I will implement that. But I will also like the keep giving the targets monitors directly. As the way I am using it is swapping it with one keybind regardless which monitors I am on. Is that fine?

packages/wm-common/src/app_command.rs
```diff
@@ -301,16 +302,36 @@ pub struct InvokeAdjustBordersCommand {
 }
 
 #[derive(Args, Clone, Debug, PartialEq, Serialize)]
-#[group(required = true, multiple = false)]
+#[group(required = false, multiple = true)]
 #[allow(clippy::struct_excessive_bools)]
 pub struct InvokeFocusCommand {
```
Im-shamo, line 307: I refactored the argument group to add the new `--summon-to-current-monitor`. This is to make sure that it the checks number of options are still there.

Im-shamo:
I ended up changing the `InvokeFocusCommand` to add the `--summon-to-current-monitor`. As this is the only why i found to do it while keeping the argument input checks. No using more then 1 command in the same time.

packages/wm/src/commands/workspace/focus_workspace_on_current_monitor.rs
```diff
@@ -0,0 +1,264 @@
...
+/// If the target workspace and focused workspace is in the same monitor,
+/// does the same thing as `focus_workspace` function.
+///
+/// If the target workspace is on a different monitor,
+/// it will move target workspace to the focused monitor and focuses it.
+///
+/// If the target workspace is displayed on a different monitor,
+/// it will swap the target workspace and the focused workspace and focuse
+/// the target workspace.
+pub fn focus_workspace_on_current_monitor(
```
lars-berger, line 28: the logic for moving workspaces is almost the same as `move_workspace_in_direction.rs`. we could change `move_workspace_in_direction` to be:

```rust
// move_workspace_to_monitor.rs
pub fn move_workspace_to_monitor(
  workspace: Workspace,
  target: MonitorTarget,
  state: &mut WmState,
  config: &UserConfig,
) -> ...

pub enum MonitorTarget {
  Index(usize)
  Direction(Direction)
}
```

with that change, we could then simply add `summon_to_current_monitor` as a new flag to the existing `focus_workspace.rs`. so if `summon_to_current_monitor` is true, then we just call `move_workspace_to_monitor(workspace, <focused_monitor>, ...)`.

lars-berger (changes requested):

lars-berger:
> This should be done.

It's looking really good 👍 These are pretty big changes, and the public api (except for some small renames) + logic look good now. I think we can simplify the code a lot though, so left a couple comments on how we can make things a lil more concise and reusable

packages/wm-common/src/app_command.rs
```diff
@@ -329,6 +350,39 @@ pub struct InvokeFocusCommand {
...
+#[allow(clippy::struct_excessive_bools)]
+pub struct SwapWorkspaceCommand {
+  #[clap(long)]
+  pub workspace_in_direction: Option<Direction>,
+
+  #[clap(long)]
+  pub monitor: Option<usize>,
+
+  #[clap(long, num_args = 2)]
+  pub monitors: Option<Vec<usize>>,
```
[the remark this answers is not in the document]
Im-shamo: This approach technically works but it cannot be called/set from the config and the id's for monitor change every time a reload.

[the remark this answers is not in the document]
Im-shamo: I will still go with the approach but that just my throughts

Im-shamo:
Done

packages/wm-common/src/app_command.rs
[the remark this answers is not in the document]
lars-berger: Yeah exactly. The long term solution imo would be having a better way to target the subject container, e.g. `glazewm c swap-workspace --monitor 0 --subject-monitor 1` (and group `subject_container_id` and `subject_monitor` together as alternatives with `clap`). Cause then we could apply this for all commands where similar functionality would be useful instead of a fix specific to `swap-workspace`

packages/wm/src/commands/workspace/swap_workspace.rs
```diff
@@ -0,0 +1,182 @@
...
+    MonitorTarget::Direction(direction) => {
+      state.monitor_in_direction(&subject_monitor, &direction)?
+    }
+    MonitorTarget::Index(index) => {
+      let monitors = state.monitors();
+      monitors.get(index).cloned()
+    }
+    MonitorTarget::Monitor(monitor) => Some(monitor),
+  };
+
```
lars-berger, line 35: Could be significantly simplified if we used the new `move_workspace_to_monitor` fn here to move the 2 workspaces between the monitors

packages/wm/src/commands/workspace/focus_workspace.rs
```diff
@@ -46,48 +48,118 @@ pub fn focus_workspace(
   }?;
 
   if let Some(target_workspace) = target_workspace {
-    info!("Focusing workspace: {target_workspace}");
```
lars-berger, line 49: I might be wrong here but can't we keep the original `focus_workspace` fn the exact same and just add to the end of it:

```rust
if summon_to_current_monitor {
  move_workspace_to_monitor(target_workspace, target_monitor)?;
}
```

lars-berger (changes requested):

lars-berger:
All the renaming changes + adding `MonitorTarget` and `move_workspace_to_monitor` look good 👍 Put in a couple comments for cleaning up the `focus_workspace` and `swap_workspace` fns

packages/wm/src/commands/workspace/move_workspace_to_monitor.rs
```diff
@@ -47,11 +57,21 @@ pub fn move_workspace_in_direction(
       );
     }
 
+    // Make sure that it is focused after move
```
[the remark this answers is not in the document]
Im-shamo: Agree

packages/wm/src/commands/workspace/focus_workspace.rs
Im-shamo: Yes.

Im-shamo:
Much thanks for the review 🙏. I am still new to programming / open source.

lars-berger (changes requested):

lars-berger:
The code for the `swap_workspace` and the new summon flag both look really nice now 👍 Saw a couple small bugs but otherwise everything is good to go. Well done!

packages/wm/src/commands/workspace/focus_workspace.rs
Im-shamo: This very broken and can not be done this way as `destroy workspaces` makes empty workspace, eg without any windows break. Moreover, this applies to `swap_workspace` fn too

Im-shamo:
> The code for the `swap_workspace` and the new summon flag both look really nice now 👍 Saw a couple small bugs but otherwise everything is good to go. Well done!

Arrr i revert the changes. Cause using `move_workspace_to_monitor` fn cause a lot errors in empty workspaces.

packages/wm/src/commands/workspace/swap_workspace.rs
Im-shamo: That does not work as well.

Im-shamo:
Thanks for all the effort. :)

Issue #973, opened 2025-02-12
[Feature Request] Add more keybind options for multiple monitor setups


nullbyto:
### Describe the problem/motivation

1. Couldn't find an option to change focus to a monitor in specific direction. Only a workaround that works in 2 monitor setup: `focus --monitor 0/1`. Otherwise i have to go through all windows in the workspaces in the direction of the second monitor to reach it.
2. Moving a window to another monitor has a similar issue, i have to move it through all other windows in the direction to reach the monitor.  I couldn't find a workaround for this.

### Describe the solution you'd like

1. Like `move-workspace --direction left/right`  there should be something like `focus-monitor --direction left/right/up/down`. Or add option to be able to specify direction within the monitor switch like `focus --monitor left/right/up/down` or add another extra switch like `--monitor-direction`.
2. Again like `move-workspace --direction left/right` there should be something to move the window to a monitor in a specific direction. `move-window --direction left/right`. Or add new option within the move command like: `move --monitor left/right/up/down` .

Im-shamo:
All of them should be done! :)
Waiting for them to get merged.

nullbyto:
Nice job! Thank you. I will try it out

PR #981, opened 2025-02-22
feat: add `--workspace-in-direction` to `focus` command

Im-shamo:
Address point 1 of #973


packages/wm-common/src/app_command.rs
```diff
@@ -307,6 +307,9 @@ pub struct InvokeFocusCommand {
   #[clap(long)]
   pub direction: Option<Direction>,
 
+  #[clap(long)]
+  pub monitor_direction: Option<Direction>,
```
lars-berger, line 311: We actually already happen to have this feature written up but it was never publicly exposed for whatever reason. There's a `WorkspaceTarget::direction` option that gets the monitor/workspace in a given direction.

Would you mind changing it to use this `WorkspaceTarget` option and exposing it as `pub workspace_in_direction: Option<Direction>`?

lars-berger (changes requested):

packages/wm-common/src/app_command.rs
Im-shamo: I see.

lars-berger (approved):

PR #989, opened 2025-02-27
feat: add monitor assignment persistence during system sleep/resume

### LINKED ISSUE #896: [Feature Request] When Connecting / Disconnecting monitors, remember previous Workspace layouts

Related to: https://github.com/glzr-io/glazewm/issues/318

I often need to connect my laptop to a Dock, work for a while with external monitors, then disconnect and work on my laptop with just the one monitor. Whenever re-connect the monitor, I have to re-configure the exact workspace layout I had before I disconnected everything. I do this quite frequently for my workflows.

1. Create a feature flag / setting to enable/disable remembering previous monitor layouts
2. If enabled, whenever you plug back into the monitors, Glaze will attempt to move Workspaces to the layout of the connected monitors previously done

### Alternatives considered

I can just reconfigure my workspace layouts everytime I re-connect and disconnect. But doing so multiple times a day + dealing with the related issue is making Glaze a friction point in my workflow.

al3rez:
Resolves #896

### DIFF
--- a/packages/wm-platform/src/event_listener.rs
+++ b/packages/wm-platform/src/event_listener.rs
@@ -10,6 +10,7 @@ pub enum PlatformEvent {
   DisplaySettingsChanged,
   KeybindingTriggered(KeybindingConfig),
   MouseMove(MouseMoveEvent),
+  SystemSleep,
   WindowDestroyed(NativeWindow),
   WindowFocused(NativeWindow),
   WindowHidden(NativeWindow),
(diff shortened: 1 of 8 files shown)


lars-berger:
Thank you for the PR! 🙌 Mentioned it on Discord, but our current approach ignores display changes during sleep, which should achieve the same result as storing/restoring monitor state on hibernate. If this PR does fix the issue, it's likely due to side effects rather than the monitor state feature itself - e.g., I notice it would skip the first display settings event after standby, which might be the actual solution. If that's the case, would you consider submitting a focused PR for just that approach instead?

PR #990, opened 2025-02-28
feat: add `--next-active-workspace-on-monitor` and `--prev-active-workspace-on-monitor` to `focus` and `move` commands

### LINKED ISSUE #938: [Feature Request] - Cycle between workspaces

Would love to have a `command` to cycle between workspaces (like left or right side of current workspace monitor) on focused monitor, instead of 'jumping' into a workspace that might belong to other monitor.

Right now I used a custom AHK script that able to cycle/move window between workspaces on each of my monitors, but would love to see it is baked in out of the box, and also would love to see the executing `command` through CLI work as well.

https://github.com/user-attachments/assets/d9ab5c7a-6791-4ff9-9496-e1798ffab8e3

### Alternatives considered

_No response_

Im-shamo:
close #938 
Also added `NextActiveInMonitor` and `PreviousActiveInMonitor` to `WorkspaceTarget` and their respective implementations.

### DIFF
--- a/packages/wm-common/src/app_command.rs
+++ b/packages/wm-common/src/app_command.rs
@@ -363,6 +363,12 @@ pub struct InvokeMoveCommand {
   pub prev_workspace: bool,
 
   #[clap(long)]
+  pub next_active_workspace_on_monitor: bool,
+
+  #[clap(long)]
+  pub prev_active_workspace_on_monitor: bool,
+
+  #[clap(long)]
   pub recent_workspace: bool,
 }
 
--- a/packages/wm-common/src/app_command.rs
+++ b/packages/wm-common/src/app_command.rs
@@ -329,6 +329,12 @@ pub struct InvokeFocusCommand {
   pub prev_workspace: bool,
 
   #[clap(long)]
+  pub next_active_workspace_in_monitor: bool,
+
+  #[clap(long)]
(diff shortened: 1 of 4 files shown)


Im-shamo:
Can also add the same option to `move` as well since the backend is in place

packages/wm-common/src/app_command.rs
```diff
@@ -328,6 +328,12 @@ pub struct InvokeFocusCommand {
   #[clap(long)]
   pub prev_workspace: bool,
 
+  #[clap(long)]
+  pub next_active_workspace_in_monitor: bool,
```
lars-berger, line 332: nit:
`next_active_workspace_in_monitor` -> `next_active_workspace_on_monitor`
`prev_active_workspace_in_monitor` -> `prev_active_workspace_on_monitor`

lars-berger (changes requested):

lars-berger:
Ay thanks for all these utility commands 🙏 They help fill in the gaps with the wm commands

> Can also add the same option to `move` as well since the backend is in place

Yeah good idea. Up to you whether you want to just include it in this PR or a new one afterwards 👍

lars-berger (approved):

lars-berger:
Perfect! Thank you 💜

Issue #776, opened 2024-10-07
[Feature Request] Rename workspace display_name


DrHeinzDoofenshmirtz:
### Describe the problem/motivation

As I see it, there is currently not an option to rename a workspace on the fly.

### Describe the solution you'd like

It would be really cool, if there was an option to rename the display_name of a workspace on the fly. It should not affect the config. Just a renaming for the current session, until config is reloaded.

PR #1042, opened 2025-04-23
feat: add `rename-workspace` command

mzhu-zx:
It seems that there isn't a command to rename workspace yet... So, this commit adds a simple command `rename-workspace`. 

Usage: 

- Using GlazeWM's term: 
  - `rename-workspace --name 1 --display-name hello`
  - `rename-workspace --name 1`: rename to "1" without a display name
  - `rename-workspace --display-name hello`: set the display name to "hello"
- It also supports a simpler i3-ish way to rename:
  - `rename-workspace 1:hello` == `rename-workspace --name 1 --display-name hello`
  - `rename-workspace 1` == `rename-workspace --name 1`
- To rename another workspace that isn't focused, specify the workspace by its name: 
  `rename-workspace --workspace 1 1:hello`
- Since a name doesn't need to be a number, the following are valid 
  ```
  rename-workspace hello:world
  rename-workspace --workspace hello 2:hello-world
  ```

Related: #776


SimpsonGSD:
This would be great to have!

lars-berger:
Thanks for working on this! 💜 I have a couple of suggestions to consider:

* Instead of `rename-workspace`, what do you think about making this a more general `update-workspace-config` command? This would let us modify other workspace options (like `keep_alive`) without needing separate commands for each property. Your structure with the clap args looks really nice, we could keep it the exact same way and just add extra optional args for the other properties:
  ```
  update-workspace-config --name <name> --display-name <display-name> --keep-alive <bool> ... --workspace <name>
  ```
* a bit of a nitpick, but I reckon requiring explicit `--name` and `--display-name` args is a bit clearer than also allowing the shorthand version, especially with the proposed `update-workspace-config` change.

DrHeinzDoofenshmirtz:
I would really like an option to rename to something on the fly, so that it is not hard-coded in the config file, but something you can do just after you have thrown a window into that workspace.

mzhu-zx:
Hi @lars-berger. Again, thanks for the suggestion and review. 

This has been done. 

```
Usage: glazewm.exe command update-workspace-config [OPTIONS] <--name <NAME>|--display-name <DISPLAY_NAME>|--no-display-name|--bind-to-monitor <BIND_TO_MONITOR>|--keep-alive <KEEP_ALIVE>>

Options:
      --workspace <WORKSPACE>
      --name <NAME>
      --display-name <DISPLAY_NAME>
      --no-display-name
      --bind-to-monitor <BIND_TO_MONITOR>
      --keep-alive <KEEP_ALIVE>            [possible values: true, false]
  -h, --help                               Print help
```

Let me know if you think there should be another improvement.

mzhu-zx:
Since the `rename-workspace` has been superseded by the new `update-workspace-config` command, the personally favored shorthand can be found in this [WinForms script](https://gist.github.com/mzhu-zx/d3710040944e425dcb2fc6ceddb77a2f) where the UI is made by AI.

lars-berger:
thanks for getting around to revamping this into the `update-workspace-config` command. the new changes look awesome! went ahead and merged 👍

PR #1047, opened 2025-04-30
fix: broadcast focus change events when focus is manually changed

dillontkh:
Before: 

https://github.com/user-attachments/assets/601242bc-0047-442a-9796-570d8357b109

After:

https://github.com/user-attachments/assets/a9068f34-89a9-4ba6-9bbe-7dbc867a5d84

### DIFF
--- a/packages/wm/src/events/handle_window_focused.rs
+++ b/packages/wm/src/events/handle_window_focused.rs
@@ -1,6 +1,6 @@
 use anyhow::Context;
 use tracing::info;
-use wm_common::{DisplayState, WindowRuleEvent};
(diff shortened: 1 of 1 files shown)


lars-berger:
Good catch, ty for investigating this! 🙏 I hope you don't mind but added a tweak so that the IPC event is broadcasted directly, since in this case we wouldn't need the other side effects from queuing a focus change. 

Sorry about taking so long to getting around to this. I've been caught up working on a new major release of Zebar the last few months that I'm really excited to get out there. Planning to go through all the open GlazeWM PRs and knock out as many bugs as I can before doing a joint release for both projects.

PR #1109, opened 2025-06-21
refactor: add `SubEnum` and `EnumFromInner` proc macros

Miitto:
Adds two new proc macros `SubEnum` and `EnumFromInner` to generate the implementations for the specialized containers.

Also includes a framework to reduce the verbosity of parsing with `syn` for any other future proc macros to utilize.


lars-berger:
Perfect! There's a couple lint errors and then it's good to merge 👌

Miitto:
Used dead code allows for the util types that I know are useful, and are being used in the other proc macro im working on. Didn't think it was worth removing them to likely just re-add them as is.

PR #1111, opened 2025-06-24
fix: window size on multiple monitors with different scaling

### LINKED ISSUE #661: Display scaling causes wrong window size
When using two monitors with different scales (e.g. 150% and 100%) under `Settings > System > Display > Scale & Layout` the size of windows is wrong. 
There are two different ways the window size is wrong. All windows are effected like described in [Problem 1](#problem1) and some are effected as described in [Problem 2](#problem2).

# Problem 1
Normal windows only have the wrong size when they are created on a secondary monitor or moved to the primary monitor. After switching workspaces the size is corrected.

# Problem 2
The size of windows like  Task Manager, Registry or Task Schedular is consistently wrong. First thought that it has something to do with admin windows but normal applications that are executed as administrator are only effected as described in [Problem 1](#problem1) and there are other applications like GIMP (v20.10.32) that are effected even though not run as admin.

For this bug to occur the displays have to overlap in width or height (see image 1).
|Problem|
|-|
|Image 1: Problem|

When they are corner to corner the problem does not occur (see image 2).
|NoProblem|
|-|
|Image 2: No Problem|

Furthermore the size is only wrong when the effected windows are at the border between the two monitors. 
Considering the configuration shown in image 1. When having two programs open next two each other, one effected (e.g. Task Manager) and one that is not effected (e.g. Command Prompt) the error only occurs when the effected window is on the right on the left monitor or on the left on the right monitor.

# Example 1
Monitor 1: 100%
Monitor 2: 150%

Admin windows on monitor 2 have wrong height.
Normal windows on monitor 2 are 150% of the right height.

# Example 1
Monitor 1: 150%
Monitor 2: 100%

Admin windows on monitor 1 and 2 have wrong size.
Normal windows have 2/3 of the right height.

I did some experimenting in order to understand the problem a bit better and gained the above mentioned information.
I hope it can help someone.
A thing that I have not really tested is whether the monitor resolution has an effect on this (I have used different resolutions).

ayushrawat10:
closes #661


ash-impl:
Screenshots for the fixes below. Note that these screenshots go across monitors with scaling 125% for left monitor and 100% for my right monitor which is the boundary crossed in these screenshots. For reference-The power icon is the rightmost icon on my YASB status bar on both monitors.

Problem 1 as stated in #661 
Before Fix
Screenshot 2025-06-25 072938
After Fix
Screenshot 2025-06-25 073245WithFix

Problem 2 as stated in #661 
Before Fix
Screenshot 2025-06-25 072746
After Fix
Screenshot 2025-06-25 073141WithFix

packages/wm/src/commands/general/platform_sync.rs
```diff
@@ -257,9 +257,21 @@ fn redraw_containers(
...
-      .apply_delta(&window.total_border_delta()?, None);
+      .apply_delta(&window.total_border_delta()?, scale_factor);
+
+    // Constrain tiling windows to their workspace's bounds to prevent
+    // overflow on multi-monitor setup with different scaling factors.
+    if window.is_tiling_window() {
+      if let Some(workspace) = window.workspace() {
+        rect = rect.constrain_within(&workspace.to_rect()?);
+      }
+    }
```
lars-berger, line 274:
```suggestion
    if window.is_tiling_window() {
      rect = rect.constrain_within(&workspace.to_rect()?);
    }
```
window's workspace is already in scope 👍

packages/wm-common/src/rect.rs
```diff
@@ -119,6 +119,25 @@ impl Rect {
     )
   }
 
+  /// Contrains this rectangle to fit entirely within the given outer
+  /// rectangle. Unlike `clamp`, this method adjusts both the position and
+  /// size to ensure the rectangle doesn't extend beyond the boundaries.
+  #[must_use]
+  pub fn constrain_within(&self, outer_rect: &Rect) -> Self {
```
lars-berger, line 126: it'd be fine to replace the `clamp` implementation with this, just need to tweak the doc comment. we only have one other use of `clamp` and there we'd want the same behavior as well

lars-berger:
thank you for the fix 💜 

do the windows in question have a very large shadow borders which then crosses over to the other monitor? i'm still not fully wrapping my head around how the problem occurs in the first place. do these wonky scaled windows take up more % than they should when you have many windows open (even after the PR fix)?

a-rawat10:
Hi @lars-berger, thank you for your response. I think the issue was more that we were passing 'None' to methods like 'apply_delta' that accepted optional scale factors for different monitors which resulted in incorrect calculations affecting window size accuracy.

ash-impl:
Hi @lars-berger, thank you for your response. I think the issue was more that we were passing 'None' to methods like 'apply_delta' that accepted optional scale factors for different monitors which resulted in incorrect calculations affecting window size accuracy.

lars-berger:
@ash-impl did some testing on a fresh W11 machine and from what i can tell, we shouldn't be scaling the border gaps. we get the border size by diffing `GetWindowRect` (includes shadow borders) and `DwmGetWindowAttribute(..., DWMWA_EXTENDED_FRAME_BOUNDS, ...)` (doesn't include shadow borders), and apparently both give already scaled rects.

<p align="center">
image
image
</p>
<p align="center">
  <i>1) with PR changes (borders are a bit wonky), 2) without changes </i>
</p>

unfortunately i'm also unable to reproduce the underlying issue (maybe it's only on W10?). monitor 1 is 150% and monitor 2 is 100%, and then i'm moving an admin window (Event Viewer) between the two monitors like in #661.

<p align="center">
image
</p>

my guess is that the root cause is `border_delta` or `shadow_border_delta` is returning some crazy large value. could you try logging them? do they return a different value when doing a cross-monitor move vs. a move within the same monitor?

```rust
// window_getters.rs

  fn total_border_delta(
    &self,
    scale_factor: Option<f32>,
  ) -> anyhow::Result<RectDelta> {
    let border_delta = self.border_delta();
    let shadow_border_delta = self.native().shadow_border_delta()?;
    tracing::info!("{:?}", border_delta);
    tracing::info!("{:?}", shadow_border_delta);
    // ...
  }
```

could you also see if it makes a difference to refresh border delta prior to redrawing?

```rust
// platform_sync.rs

    window.native().refresh_border_position()?; // add this
    
    let rect = window
      .to_rect()?
      .apply_delta(&window.total_border_delta()?, None); // leave the rect calculation as-is
```

ash-impl:
Hi @lars-berger thank you for your response. 
You're right about not wanting to scale border gaps.

I have since switched to using a single monitor only thanks to your great work on this application and then workplace upgraded to W11 (where I'm not sure this issue persists because of single monitor use).
I saw the issue only with Windows applications like Event Viewer, task manager etc. Other applications were working just fine. I have also since then gained a better understanding of this code-base for personal learning. And I can't seem to pin point why those applications in particular were the ones giving grief. My initial changes seemed to have worked for me in solving those issues with these code changes back when I was working on this.

I can test with your recommendations of logging, and work on this more in a couple weeks if it's still an issue (On W10 or W11).

PR #1121, opened 2025-06-28
fix: avoid recentering manually placed floating windows when unplugging usbs

Ciuriya:
Currently in GlazeWM, it assumes any change reported by the OS is a display setting change, but sometimes this is just an unrelated event, such a usb being unplugged/replugged.

The problem is that if the user has a floating window that has been positioned (either manually moved or moved via the position command), this window will be automatically recentered because the positioning might have changed.

This PR checks if the window was placed by the user and if it still overlaps with the workspace after the change before forcefully recentering.


lars-berger:
@Ciuriya thank you for the pr! 🙏  apologies for getting around to this so late.

Issue #1138, opened 2025-07-19
[Bug] can't ignore PowerToys Command Palette


chadluo:
### Describe the bug

[Command Palette](https://learn.microsoft.com/en-us/windows/powertoys/command-palette/overview) is a new launcher provided in PowerToys, similar to the existing PowerToys Run.

However under GlazeWM, Command Palette is always triggered to tiling mode unless using `Alt-T` to move it off tiles. I tried to add the name to config ignore list but doesn't seem to work:

```yaml
window_rules:
  - commands: ["ignore"]
    match:
      - window_process: { equals: "Microsoft.CmdPal.UI.exe" }
```

The name `Microsoft.CmdPal.UI.exe` is both the file name in install directory and the process name in Task Manager.

### Reproduction

1. install PowerToys
2. enable Command Palette in PowerToys
3. trigger Command Palette with shortcut

### Stack trace or error logs (if applicable)

```shell

```

### Version number

3.8.1

Lubieerror:
After fighting with WinSpy++ I think I found a solution:
```yml
      - window_title: { equals: 'Command Palette' }
        window_class: { equals: 'WinUIDesktopWin32WindowClass' }
```

chadluo:
Works for me as well 👍 

Can give WinSpy++ a try later I guess

Jus2Cat:
This tip might seem obvious to some, but if your system language is set to something other than English, you'll also need to change the `window_title` accordingly 

For example, in an English-language system:

```yaml
- window_title: { equals: 'Command Palette' }
  window_class: { equals: 'WinUIDesktopWin32WindowClass' }
```

In Russian:

```yaml
- window_title: { equals: 'Палитра команд' }
  window_class: { equals: 'WinUIDesktopWin32WindowClass' }
```

TomasMontana:
In Spanish: 
` - window_title: { equals: 'Paleta de comandos' }
    window_class: { equals: 'WinUIDesktopWin32WindowClass' } `

Notice that apparently in every language other than english (at least that seems to be the case for russian and spanish) only the first word of the window_title is capitalized.

ebrithiljonas:
Thanks for the solution, in German this works:

```yaml
- window_title: { equals: 'Befehlspalette' }
  window_class: { equals: 'WinUIDesktopWin32WindowClass' }
```

PR #1140, opened 2025-07-22
fix: Update config to ignore new PT Command Palette

Lubieerror:
Microsoft recently added a new Command Palette/Runner to PowerToys and it's getting tiled ( #1138 ).
After fighting with WinSpy++ I think I found the correct title + class combination to ignore it. Works for me, but I'm still new to both the project and "Windows windows inspections". If I messed up something else, or maybe there's a more optimal way to do this change, lmk (or take inspiration and close this commit - I'm ok with that).

Thanks for software and have a great day!


LukeDRussell:
Works for me.

dithmer:
> Works for me.

I can support that statement.

lars-berger:
thank you 🙏

Issue #668, opened 2024-08-10
floating windows are centered by default disable


xxmlud:
Hello,

I am using glazewm on Window 10.
I can't find the option to disable the option to have a window appear in the center by default when I start a window.

My configuration is as follows, but always when I open a window it starts in the center. 
I would like the behavior to be like the preview in the README.md, that the windows start to the right/left of the monitor.

```
  state_defaults:
    floating:
      # Whether to center floating windows by default.
      centered: false

``` 

Thank you

lars-berger:
Was able to reproduce this

yutotakano:
On top of floating windows starting centered, my windows sometimes suddenly become centered after launch too. I could be calculating something with the calculator app or writing in notepad.exe (both floating) and they'd suddenly be moved to the center of the monitor. It is very sporadic and frustratingly random, so much that I thought it might be a OOM or CPU exhaustion issue that causes GlazeWM to maybe restart, but this does not appear to be the case. So I have no idea.

The following is my relevant portion of the config. I don't have any keybindings to make a window centered either, so there's no accidental presses of something.

```
window_behavior:
  initial_state: 'floating'

  state_defaults:
    floating:
      centered: false
```

alextrepa:
I also have this problem. This sadly makes it unusable for me. If it helps, I noticed it does that only when an IDE from Jetbrains is open (PhpStorm, Datagrip, Webstorm, etc.). The IDE is also flickering a bit. I know nothing of tile management, but to me it looks like something is interfering with the rendering of the window. 

I don't have any video, but if that helps it's very simlar to this issue on Komorebi: https://github.com/LGUG2Z/komorebi/issues/781. If you want to tackle this bug and need more info, I could try and reproduce it.

mheraskin:
@lars-berger 
I have all the floating windows randomly getting centered in the middle of work, so I have to rearrange them again. It's really frustrating.

babysnakes:
Hi, I've only started to use glazewm a few days ago and I'm loving it. 

[...]

PR #1166, opened 2025-09-04
Optionally prevent centering of floating windows on display changes.

babysnakes:
Fix: #668.

> Note: the PR origin is from the stable TAG. I didn't know if the later commit is stable and I needed something I can trust that it's working. If needed I'll merge _main_ into it. This is why it's a Draft PR.

### Motivation:

I've started using GlazeWM a couple of weeks ago. I was looking for a way to use mostly floating windows, but for specific cases, I wanted to be able to tile the windows and get the full experience of a tiling window manager. GlazeWM exceeded my expectations. One big problem was that maybe because of my USB switcher or my keyboard, several times a day all my floating windows got centered. Every time the computer went to sleep, or I would switch to my work laptop, or even just because :slightly_frowning_face:, All my windows would become centered. This really disturbed my workflow for such extent that I decided to look for a solution. 

Please note that while I have some experience with Rust, I have zero experience in desktop programing and architecture so I may have missed many things completely.

### My suggestion:

 I checked two areas of the code. The `event_window_proc` function where events are filtered and the `handle_display_settings_changed` where the windows get centered. I printed out all the events while connecting/disconnecting devices and monitors, but while I could reduce the number of windows centering occurrences if I disabled `WM_DEVICECHANGE`, it didn't completely stop the false identification of display changes, so I figured out I need to look at the 'handle_display_settings_changed' function. Here's the [code](https://github.com/glzr-io/glazewm/blob/6242f8a0c5754007c6b286bcf6a555f328c1c829/packages/wm/src/events/handle_display_settings_changed.rs#L119-L128):

```rust
    // Need to update floating position of moved windows when a monitor is
    // disconnected or if the primary display is changed. The primary
    // display dictates the position of 0,0.
    let workspace = window.workspace().context("No workspace.")?;
    window.set_floating_placement(
      window
        .floating_placement()
        .translate_to_center(&workspace.to_rect()?),
    );
  }
```
It seems to me that it automatically centers the floating position of all windows in `state.windows()` (which in my tests result in all windows, not just the moved ones (and I wrong?) - the comment may be a little misleading). Since I'm a new user of GlazeWM, I trust that this is for a good reason, However, I wanted to allow users to disable this behavior at the possible price of having windows be partly off-screen if disconnecting from a higher resolution monitor.

I've added a new config option `prevent_centering_of_floating_windows` which default to `false` (current behavior) but when set to true, tries to leave windows where they are and move only windows that are completely off-screen.

If this is acceptable, I also want to extract two functions from the `handle_display_settings_changed` function (one that handles state workspace updates and the second that handles the windows) because it seems that they are independent of each other.

One last note, there's always the option to do a much deeper change and add to the windows metadata the `Rect` relation to the parent workspace so a window will not get moved unless it really changed to a monitor with different size, but I think the current suggestion is probably good enough to start with. 

And thanks again for this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,13 @@ general:
   # workspace when focusing the current workspace.
   toggle_workspace_on_refocus: false
 
+  # GlazeWM, by default, centers all floating windows when
+  # connecting/disconnecting monitors. When working with many floating windows,
+  # this could disturb your workflow. Setting this to true will try to leave
+  # floating windows at their current position at the price of having windows
+  # partly off the screen when switching to a lower resolution monitor.
+  prevent_centering_of_floating_windows: false
+
   cursor_jump:
     # Whether to automatically move the cursor on the specified trigger.
     enabled: true
--- a/packages/wm-common/src/parsed_config.rs
+++ b/packages/wm-common/src/parsed_config.rs
@@ -93,6 +93,10 @@ pub struct GeneralConfig {
 
   /// Affects which windows get shown in the native Windows taskbar.
   pub show_all_in_taskbar: bool,
+
+  /// Whether to prevent glazewm from centering the floating placement of all
+  /// windows in case of suspected monitor change.
+  pub prevent_centering_of_floating_windows: bool,
 }
 
 impl Default for GeneralConfig {
@@ -106,6 +110,7 @@ impl Default for GeneralConfig {
       config_reload_commands: vec![],
       hide_method: HideMethod::Cloak,
       show_all_in_taskbar: false,
+      prevent_centering_of_floating_windows: false,
     }
   }
 }
--- a/packages/wm/src/events/handle_display_settings_changed.rs
+++ b/packages/wm/src/events/handle_display_settings_changed.rs
@@ -1,5 +1,6 @@
 use anyhow::Context;
 use tracing::info;
+use wm_common::Rect;
 use wm_platform::Platform;
 
 use crate::{
@@ -120,11 +121,17 @@ pub fn handle_display_settings_changed(
     // disconnected or if the primary display is changed. The primary
     // display dictates the position of 0,0.
     let workspace = window.workspace().context("No workspace.")?;
-    window.set_floating_placement(
-      window
-        .floating_placement()
-        .translate_to_center(&workspace.to_rect()?),
-    );
+    let workspace_rect = workspace.to_rect()?;
+    let floating_placement = window.floating_placement();
+    let new_placement =
+      if config.value.general.prevent_centering_of_floating_windows {
+        only_move_off_screen_windows(&workspace_rect, &floating_placement)
+      } else {
+        Some(floating_placement.translate_to_center(&workspace_rect))
+      };
+    if let Some(placement) = new_placement {
+      window.set_floating_placement(placement);
+    }
   }
 
   // Redraw full container tree.
@@ -133,4 +140,17 @@ pub fn handle_display_settings_changed(
     .queue_container_to_redraw(state.root_container.clone());
 
   Ok(())
+}
+
+fn only_move_off_screen_windows(workspace_rect: &Rect, floating_placement: &Rect) -> Option<Rect> {
+  if floating_placement.x() > workspace_rect.width()
+      || floating_placement.y() > workspace_rect.height()
+  {
+    Some(floating_placement.translate_to_coordinates(
+      floating_placement.x() % workspace_rect.width(),
+      floating_placement.y() % workspace_rect.height(),
+    ))
+  } else {
+    None
+  }
 }
--- a/resources/assets/sample-config.yaml
+++ b/resources/assets/sample-config.yaml
@@ -18,6 +18,13 @@ general:
   # workspace when focusing the current workspace.
   toggle_workspace_on_refocus: false
 
+  # GlazeWM, by default, centers all floating windows when
+  # connecting/disconnecting monitors. When working with many floating windows,
+  # this could disturb your workflow. Setting this to true will try to leave
+  # floating windows at their current position at the price of having windows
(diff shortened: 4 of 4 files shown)


babysnakes:
I've been using it for about a month and it seems to be working correctly.  I still offer the option to extract the `handle_display_settings_changed ` function to two sub functions to make it clearer (while I'm already working on this code).

Thanks

dizzi1222:
Pls need to fix it, when I take out the USB it focus center on floating mode.

babysnakes:
It seems that #1121 deals with the same issue. While I prefer the more global solution suggested here, I admit that my use case is not the default one (I use floating windows by default and with the merged solution I'll have to reposition each new window - annoying but I can live with it).

@lars-berger, please let me know if you prefer the merged solution or this one.

lars-berger:
@babysnakes ah interesting, is it due to [this added section](https://github.com/Ciuriya/glazewm/blob/1c9856e407e89f4509c6e7828de75f9face74f7b/packages/wm/src/commands/window/set_window_position.rs#L39-L42) that floating windows then have to be repositioned manually on open?

IMO it makes most sense not to have a config option for whether to recenter, since many users view it as a bug that non-display changes (e.g. unplugging USBs) can trigger recentering. Would it solve your new issue if we changed how `has_custom_floating_placement` gets updated?

babysnakes:
@lars-berger, thanks for your response.

> @babysnakes ah interesting, is it due to [this added section](https://github.com/Ciuriya/glazewm/blob/1c9856e407e89f4509c6e7828de75f9face74f7b/packages/wm/src/commands/window/set_window_position.rs#L39-L42) that floating windows then have to be repositioned manually on open?

Probably, I actually read the description, not the entire PR code (I only checked for the collision with my PR).

I don't disagree regarding the configuration. I only suggested it because I assumed most users are ok with windows getting centered. How do you suggest `has_custom_floating_placement` should get updated? Currently its false by default, and only if you move the window, it changes to `true`. If you want to set it to true by default, you practically disable the centering unless you add a command to manually set centering mode on a window. I would love to have windows not centered by default.

Thanks

lars-berger:
@babysnakes no problem! i guess that means your config looks like this?
```yaml
window_behavior:
  initial_state: "floating"
  state_defaults:
    floating:
      centered: false
```
on a Mac laptop atm - could you see if changing [this line (L#163 of `manage_window.rs`)](https://github.com/glzr-io/glazewm/blob/main/packages/wm/src/commands/window/manage_window.rs#L163) fixes the issue you mentioned about repositioning new windows?
```diff
-   false,
+  !prefers_centered,
```

babysnakes:
Thanks @lars-berger. 

Regarding your suggested change, in the little testing I did (and without inspecting the code), it seems to not center newly created windows. I certainly prefer this option 🙂.

I did however, find unexpected behavior (even before the change you suggested). Let me try to describe, if it's not clear I can film it with my mobile and upload.

I have a laptop with two screens. When I first tested main after the merge of #1121, I had on my first screen:
- Couple of windows in default location
- Couple of moved windows

I then disabled / enabled the second screen, and it behaved as expected (center only the unmoved windows).

However, now I did some more tests: 
- On the first screen I had exactly the same as above
- On the second screen I had a single window which I moved manually

After disabling the second screen I saw the following behavior (both with and without your suggested change):
- The windows on the first screen behaved as above
- The moved window on the second screen was centered (on workspace "2")

Is this expected behavior? (in my code I handled this differently - I only moved windows that were out of the workspace rectangle).

I can try next week to take a deeper look and see if I find the reason for it.

lars-berger:
> After disabling the second screen I saw the following behavior (both with and without your suggested change):
> 
> * The windows on the first screen behaved as above
> * The moved window on the second screen was centered (on workspace "2")
> 
> Is this expected behavior? (in my code I handled this differently - I only moved windows that were out of the workspace rectangle).

So workspace 2 was moved from monitor 2 (now disabled) -> 1 and then its windows were centered after the move? That sounds correct to me - we do the same when moving workspaces via `alt+shift+<asdf>`. The thought behind it is that monitors can be different sizes, so floating windows might be placed into an awkward spot even though they still fit within the rect of the new monitor (especially if you have monitors that are mixed landscape + portrait).

I can see how that'd be annoying for users who float every window though 😕. For your use case, would it be a better experience if they were moved without centering?

babysnakes:
> I can see how that'd be annoying for users who float every window though 😕. For your use case, would it be a better experience if they were moved without centering?

I see about the two monitors. Let me try to summarize. There are two issues:
1. General floating windows centering on USB connect/disconnect.
2. Centering floating windows when disconnecting monitors

### Centering windows on USB connect/disconnect

The solution in #1121 is an elegant one,. However, if, as you said, most users treat windows centering as a bug, maybe we should disable windows centering on USB connect/disconnect all together. Settings the default `has_custom_floating_placement` to true, makes this property irrelevant unless you add a command to force window centering (not sure it makes much sense).

### Centering floating windows on monitor disconnect

This is actually not a big problem for me. 99% of my time is working on my PC or my laptop using single monitor. Even when I use a second monitor, it's mostly for fullscreen window. If I'm not the only one that complained about it, I can suggest a setting that instead of centering the windows. just make sure they are within the viewable area at the price of occasionally being placed awkwardly.

**Edit**: @lars-berger , on second thought, It might be a good idea to add it as setting. I think that if I was constantly connecting and disconnecting my laptop to a monitor, I personally would be more annoyed by all the windows getting centered than by occasional awkward placement. It would be easy to update my PR to handle just this case.

#### Summary
I'll be happy to implement any change you want to make (if any). For me the only thing that really matters is to minimize windows centering as much as possible.

Thanks

babysnakes:
Created #1304 as an alternative.

PR #1182, opened 2025-09-22
feat: add support for `henkan` and `muhenkan` keys on JP keyboards

### LINKED ISSUE #1112: [Feature Request] Key request

I want to use muhenkan and henkan key in glazewm.
could you add them?
VK_NONCONVERT
VK_CONVERT

### Alternatives considered

_No response_

tanitta:
This PR adds support for the `NonConvert` (無変換) and `Convert` (変換) keys 
on Japanese (JP) layout keyboards.

On JP layout keyboards, users of tiling window managers often remap the 
`NonConvert` and `Convert` keys as convenient modifier keys ( close #1112  ).


lars-berger:
thanks for the pr! 🙏

renamed it to "henkan" and "muhenkan" since that seems a bit more common ([and it's apparently what other platforms use to refer to those keys](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values#common_ime_keys))

PR #1184, opened 2025-09-25
fix: "No displayed workspace" after waking up from sleep

vantm:
Closes #907
Closes #1124

After the computer wakes from sleep or reconnects to a monitor, Glaze tries to add new monitors and arrange workspaces and windows. However, there’s a statement that retrieves workspace information before the arrangement, which causes the bug. Simply move that statement to run right after the workspace has been arranged.

PR #1185, opened 2025-09-27
fix: Incorrect workspace bindings when disconecting then reconnecting a monitor

### LINKED ISSUE #1130: [Bug] inconsistancy with connecting and disconnecting external monitors
### Describe the bug

In my workflow - I have two external monitors and one laptop monitor.
If i keep the standard workspaces (9) - then after connecting/disconnecting monitors - after disconnecting external monitors - apps get spread across different workspaces. Since I do not need more than one workspace per monitor and not using Zenbar - I have no way accessing other workspaces other that using keybinding which is not desirable. What would be great - have ability to ditch workspaces completely and have one monitor = one workspace.

Solution: allow using single workspace on multiple monitors

### Reproduction

```
workspaces:
  - name: "1"
    keep_alive: true
    bind_to_monitor: 0
  - name: "2"
    keep_alive: true
    bind_to_monitor: 1
  - name: "3"
    keep_alive: true
    bind_to_monitor: 2  # Laptop
```

Then after disconnecting (then i don't see apps that were on workspace 1 and 2) and connecting back external monitors i get an error: `no displayed workspace`. All workspaces stay on the laptop screen and i would need to restart the app. 

If i use the same config but without binding to monitor, then after reconnecting external monitors i get this: `No workspace config available to activate workspace`

```shell

```

### Version number

3.9.1

vantm:
… a monitor

Close #1130

### DIFF
--- a/packages/wm/src/commands/monitor/add_monitor.rs
+++ b/packages/wm/src/commands/monitor/add_monitor.rs
@@ -17,8 +17,7 @@ use crate::{
 pub fn add_monitor(
   native_monitor: NativeMonitor,
   state: &mut WmState,
-  config: &UserConfig,
-) -> anyhow::Result<()> {
+) -> anyhow::Result<Monitor> {
   // Create `Monitor` instance. This uses the working area of the monitor
   // instead of the bounds of the display. The working area excludes
   // taskbars and other reserved display space.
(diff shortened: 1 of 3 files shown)


vantm:
Need #1184 to be merged in order to reproduce this bug.

thundron:
FWIW: tested both PRs on my machine, seems to have solved these specific multi-monitor issues (binding initially still refers to the wrong monitors, but I believe that's a different issue)

PR #1188, opened 2025-09-30
fix: add cursor jump for `wm-cycle-focus` and window minimize

### LINKED ISSUE #884: [Bug] cursor_jump does not jump when moving windows
### Describe the bug

Setting the cursor_jump trigger to window_focus does not cause the cursor to jump when moving a window around. This results in problems when attempting to use focus follows mouse, as chaining movements becomes impossible.

### Reproduction

1. set `trigger` under `general:cursor_jump` to `window_focus`
2. switch to a different a window, the cursor will jump onto the new window
3. move a window, the cursor will not follow the moved window and will instead focus on another window

_No response_

### Version number

3.7.0

Ma-Shell:
This PR closes #884 and expands it to more events.
The `cursor_jump` configuration setting was previously not triggering on various occasions. This is especially annoying when mixed with the `focus_follows_cursor`-option for options which move the position of the active window, if the cursor then automatically focuses a different window. This PR thus adds cursor jump for multiple events:
- cycling focus
- moving window in direction
- toggling tiling
- focusing floating window

### DIFF
--- a/packages/wm/src/commands/window/move_window_in_direction.rs
+++ b/packages/wm/src/commands/window/move_window_in_direction.rs
@@ -27,7 +27,6 @@ pub fn move_window_in_direction(
   state: &mut WmState,
   config: &UserConfig,
 ) -> anyhow::Result<()> {
-  state.pending_sync.queue_cursor_jump();
   match window {
     WindowContainer::TilingWindow(window) => {
       move_tiling_window(window, direction, state, config)
--- a/packages/wm/src/commands/window/update_window_state.rs
+++ b/packages/wm/src/commands/window/update_window_state.rs
@@ -111,8 +111,7 @@ fn set_tiling(
   state
     .pending_sync
     .queue_containers_to_redraw(target_parent.tiling_children())
-    .queue_workspace_to_reorder(workspace)
-    .queue_cursor_jump();
+    .queue_workspace_to_reorder(workspace);
 
   Ok(tiling_window.into())
 }
@@ -138,7 +137,6 @@ fn set_non_tiling(
     return Ok(window);
   }
 
-  state.pending_sync.queue_cursor_jump();
   let workspace = window.workspace().context("No workspace.")?;
 
   match window {
--- a/packages/wm/src/events/handle_window_minimized.rs
+++ b/packages/wm/src/events/handle_window_minimized.rs
@@ -36,7 +36,7 @@ pub fn handle_window_minimized(
       if let Some(focus_target) = state.focus_target_after_removal(&window)
       {
         set_focused_descendant(&focus_target, None);
-        state.pending_sync.queue_focus_change();
+        state.pending_sync.queue_focus_change().queue_cursor_jump();
         state.unmanaged_or_minimized_timestamp =
           Some(std::time::Instant::now());
       }


lars-berger:
thank you for the pr! :pray:

100% agree that cursor jump needs improvement for when `focus_follows_cursor` is enabled. i'm thinking we should add a new cursor jump `trigger` option though (e.g. `trigger: "window_focus_or_move"`), since `trigger: "window_focus"` doesn't make it clear that it jumps on both window focus + move (and there's still a use case for _only_ jumping on window focus). 

`cycle_focus` + `handle_window_minimized` should still be triggering a cursor jump regardless, so tweaked the pr a bit so that those changes can get merged in.

---

took a quick look at what would be needed to add `trigger: "window_focus_or_move"` and there's 2 things to account for:
* problem: if we add `queue_cursor_jump()` to a bunch of new locations, it should only queue the jump if a window that's been moved currently has focus.
  * solution 1) always check `window.has_focus` before queuing the jump.
  * solution 2) remove `queue_cursor_jump()` and instead derive whether to cursor jump from `containers_to_redraw` → if the focused window will be moved, then we should cursor jump. 
    * problem: window might be queued for redraw for only a resize, in which case, we wouldn't want to cursor jump.
      * solution 2a) replace `queue_containers_to_redraw` with `queue_window_to_move` and `queue_window_to_resize`. get whether to cursor jump based on if any of the windows to move are focused.
* problem: with both `"window_focus_or_move"` and `"window_focus"` available as triggers, we need to check in `platform_sync` if a cursor jump is queued due to window move or focus.
  * solution 1) add a new param to `queue_cursor_jump` to indicate whether it's due to window focus or move.
  * solution 2 from above works here too.

solution 2a. looks like it's a bit more robust. do you wanna take a stab at this @Ma-Shell or @gsfernandes81?

Ma-Shell:
I would argue that when resizing a cursor jump might also make sense, e.g. if the cursor is on one side of the window and the window is shrunk so that after the resize the area where the cursor is now belongs to a different window, which would then be automatically focused.

Maybe an alternative solution would be to introduce a separate command like, which the user can add to the configuration keybindings, e.g. 
```yaml
  - commands: ['move --direction left', 'cursor-jump']
    bindings: ['alt+shift+h', 'alt+shift+left']
```
While this would make the config a bit more verbose, this way the user has a fine grained control in which situations the cursor jump happens. It should be ensured though that the second command is triggered before any implicit focus changes by `focus_follows_cursor` which the first command might have caused.

The existing trigger for the cursor_jump would still make sense though for situations where the focus change wouldn't occur due to any direct command by the user (e.g. spawning / closing windows, alt+tab or some other means of focus change)

lars-berger:
> I would argue that when resizing a cursor jump might also make sense, e.g. if the cursor is on one side of the window and the window is shrunk so that after the resize the area where the cursor is now belongs to a different window, which would then be automatically focused.

true, that's a good point ➕ 

currently with `trigger: "window_focus"`, we *always* cursor jump if it's queued, even if the cursor is already over the window. the positive of this is that it makes it more predictable - e.g. with a new `trigger: "window_focus_or_move"` option, when you `alt+shift+space` to toggle floating, it'd mean your cursor is always in the center of the window. but the downside is it'd be annoying if it's jumping in scenarios when you don't expect it to. this could _potentially_ be fixed by having a config option of only jumping if the cursor is outside the window, but even then, users who have that disabled still would likely not want it to jump on small resizes.

> Maybe an alternative solution would be to introduce a separate command like, which the user can add to the configuration keybindings, e.g.
> ...

[Aerospace](https://nikitabobko.github.io/AeroSpace/guide#config-samples) actually handles it in a pretty similar way. they instead have a few scriptable event handlers that you can then use to run a cursor jump command:

```toml
# Mouse follows focus when focused monitor changes
on-focused-monitor-changed = ['move-mouse monitor-lazy-center']
```

one downside with having `['move --direction left', 'cursor-jump']` is that it wouldn't affect move commands run via IPC . it could still be worth adding a `cursor-jump` command though! super cool imo.

PR #1226, opened 2025-12-13
fix: missing `queue_cursor_jump()` when moving window to sibling containers

gsfernandes81:
This PR fixes an issue where moving a window within the same workspace does not trigger a cursor jump, even with the `cursor_jump` feature enabled.

PS: First time contributing to this project and working on any rust project in general. Please let me know in case of any issues with the PR and looking forward to this fix being merged if possible.


lars-berger:
left a comment on the PR over at https://github.com/glzr-io/glazewm/pull/1188#issuecomment-3678834047 which had a similar fix. if you're keen to take a stab at the change mentioned in the comment, it'd be best to have it in a separate PR since the approach would be quite different - i'll go ahead and close this one.

PR #1242, opened 2026-01-04
feat: Persist and auto-restore window's workspace assignments across crashes

### LINKED ISSUE #1241: [Feature Request] Persist window workspace after wm terminates/crashes

`glazewm` occasionally crashes in a non-deterministic way.
The crash reason is difficult to identify and is not always the same.

However, one behavior is consistent after a crash:
- All managed windows are stacked into a single workspace
- On restarting `glazewm`, the window manager has no memory of previous workspace assignments
- Users must **manually move each application back to its intended workspace**

For users running many applications across multiple workspaces, this becomes extremely frustrating and time-consuming. Until `glazewm` reaches a point where crashes are effectively impossible, t**he lack of state recovery makes crashes disproportionately disruptive.**

Introduce a lightweight window-to-workspace persistence mechanism that allows glazewm to restore workspace assignments after restart.

Persist a mapping of managed windows - workspaces to disk and restore it on startup.

**Runtime behavior**

- **During normal operation**
-- Enumerate currently managed windows
-- Merge / overwrite entries for active windows
-- Continuously keep the mapping file up to date

- **On startup**
-- Read the mapping file
-- Enumerate native windows
-- Match windows using available identifiers (handle, process name, class, title)
-- Reapply workspace assignments using existing manage logic

- **On manual user interaction**
-- When a window is manually focused or moved:
-- Check for a matching mapping entry
-- Apply workspace if applicable
-- Remove the entry once resolved (to avoid stale state)

This mechanism is intended as a best-effort recovery, partial matches are acceptable if exact identifiers are unavailable, stale entries can be cleaned opportunistically.

### Alternatives considered

_No response_

Arun0A:
This PR adds an **optional, crash-resilient (window, workspace) persistence mechanism** to GlazeWM.

When enabled, GlazeWM persists minimal window metadata and restores workspace assignments on restart, preventing the common post-crash scenario where all windows stack into a single workspace and must be manually reorganized.

Until crashes are fully eliminated, restoring workspace state significantly improves usability and perceived stability.
Closes [#1241](https://github.com/glzr-io/glazewm/issues/1241)

**To enable this feature put this in config under general,**
`persists_process_location = true`
_(Disabled by default)_
All persistence, restore, and cleanup logic is gated behind this flag

### How this works:
**Mapping format & storage**
JSON file: glazewm_apps_workspaces.json
Stored next to the user config (fallback: executable directory)

**Schema:**
{ handle, process_name, class_name, title, workspace }
Atomic updates via temporary file + rename to avoid corruption

**Shared type**
Introduced AppWorkspaceEntry (in wm-common/app_workspace.rs)
Used across crates to avoid duplicate structs and drift

#### Normal operation

* `platform_sync` periodically collects managed windows
* Merges / overwrites entries for active windows
* Writes updated mapping atomically (if enabled)

#### Startup auto-restore

* `WindowManager::auto_restore_mappings()` runs on startup
* Reads mapping file and enumerates native windows
* Matches windows using heuristics:

  1. Native handle (primary)
  2. Process name equality
  3. Title substring match
* Restores matched windows to their saved workspace via existing manage logic

#### On manage / focus

* **Manage-time** (`manage_window.rs`)

  * If a newly managed window matches a mapping entry:

    * Move to saved workspace
    * Remove entry
    * Write mapping atomically

* **Focus-time** (`handle_window_focused.rs`)

  * If a window is manually focused and matches:

    * Move to saved workspace
    * Remove entry

#### Cleanup

* Watcher listens for `WindowUnmanaged`
* Removes stale entries for terminated handles
* Runs only if persistence is enabled

#### CLI helper

* Added `restore-mapping` command
* Enumerates native windows and focuses unmatched ones so the WM can manage and restore them
* Useful if windows existed before WM startup or were not immediately adopted

**Additional notes:**
I had tried to identify the windows immediately after the startup to send them to their mapped workspace before focusing. However, I could not manage to do that, for some reason, the process would not be identified unless they were focused.


Louis047:
i will test your PR branch tomorrow and let you know if they have fixed the persistent workspace issue. In the meantime, I went ahead and implemented a fix for monitor disconnect/reconnect issue where it throws `No displayed workspace` issue which is a fatal error and doesn't allow to use that workspace anymore.

PR: #1268 

Update: I have closed the PR for the monitor disconnect/reconnect issue which was already address here #1185

Arun0A:
@Louis047 did you get a chance to check this?

Louis047:
No unfortunately, I was caught up with other commitments and completely forgot about this. I think this is fixed in the new update so do check it out.

Louis047:
@Arun0A i think this issue has been mitigated already in the latest update through different PRs that were made earlier which has fixed multi-monitor handling issues and the missing explorer hook after restarting explorer. 

I just tested the latest build and everything that this PR tried to address has been fixed already.

Arun0A:
ok - I will not have access to my machine for some time to test this.

however, i would like to know, with the other PRs that fixed this, does it apply to preserving the window positions/dimensions in a workspace after restart?

if not, then the approach in this PR could be usefull in preserving the states of gwm. if you still prefer not to implement this, you may close this PR.

Thanks.

Louis047:
@Arun0A Yeah, it was implemented already, I can confirm it but I kinda hate the fact that some of the crucial bugs like this: #1275 went unnoticed ever since the first release of rust rewrite.

Arun0A:
@Louis047 okay sounds good

> ... some of the crucial bugs like this: #1275 went unnoticed ...

i'll take a closer look at issue #1275, let me see if i can figure out the cause and possibly come up with a fix.

Louis047:
@Arun0A It's better if you close this PR now. It's been too long and the dev seems to be very inactive for the past two months due to his work. 

And as I said earlier. This has already been resolved.

Issue #1243, opened 2026-01-06
[Bug] Firefox behavior for web links


marcmillien:
### Describe the bug

Hello :).

This bug report concerns the firefox link behavior accross different workspaces.

With the `hide_method: cloak` opening links on workspace that has no firefox instance opens them in new firefox window.
With the legacy `hide_method: hide` it opens a new tab in the existing firefox window and automatically switches to that workspace.

As the firefox is configured to open link in new tab, I think there is an issue with the `cloak` method (I don't think automatic switch is an issue as this is not a firefox configuration).
I don't see additional parameter in firefox `about:config` that could force the correct behavior.

I create this issue, there is also a discussion https://github.com/glzr-io/glazewm/discussions/1216 from @PrimaMateria

### Reproduction

Open a web link from an other application in a different workspace using cloak hide method.

### Stack trace or error logs (if applicable)

```shell
None
```

### Version number

3.9.1

marcmillien:
I found a solution using hide method per process that seems to work if some people can also test it before I make the pull request ready :).
https://github.com/glzr-io/glazewm/pull/1248

bearylabs:
You can fix it in Firefox itself in about:config, just set widget.prefer_windows_on_current_virtual_desktop = false

jondmarien:
Confirming the same behavior with Zen Browser (Firefox-based) on Windows + hide_method: cloak:

- Chrome: clicking an external link focuses the existing window and GlazeWM switches workspace.
- Zen: opens a new window on the *current* workspace instead of activating the existing Zen window.

Workaround that worked: about:config → widget.prefer_windows_on_current_virtual_desktop = false
(Also related Mozilla bugs [1997038](https://bugzilla.mozilla.org/show_bug.cgi?id=1997038) / [1423768](https://bugzilla.mozilla.org/show_bug.cgi?id=1423768) — cloaked/other-desktop windows are skipped when remoting external URLs.)

Happy to test a cloak visibility fix or per-process hide_method if revived.

PR #1247, opened 2026-01-15
Hide method cloak should make windows API visibility aware

marcmillien:
This tries to fix firefox web link behavior across workspaces.

Cf.: https://github.com/glzr-io/glazewm/issues/1243


marcmillien:
This solution does not work, I close this PR

PR #1248, opened 2026-01-16
feat: Offer hide method per process option

marcmillien:
Adds the possibility to choose hide_method per process so firefox can act as expected (open links on tab in existing firefox window)

More details in https://github.com/glzr-io/glazewm/issues/1243

I first tried to "fix" with https://github.com/glzr-io/glazewm/pull/1247 but it does not work and make some bugs solved by cloak back, so I came to this solution.


marcmillien:
After one week of test, it does the trick although this is not perfect (bugs with hide_method are back for firefox).

marcmillien:
No return, files to rebase, I prefer to close this PR, if someone is interested, build it or update it :)

Issue #1104, opened 2025-06-18
[Feature Request] An option to exclude non-focused fullscreen windows from transparency effects


loobylooby:
### Describe the problem/motivation

When using the transparency setting for non-focused windows, fullscreen windows also become transparent when they lose focus.

### Describe the solution you'd like

If possible, introduce another configuration option that allows fullscreen windows to be excluded from the non-focused transparency effect. With this secondary option enabled, non-focused fullscreen windows would remain fully opaque, while non-focused windows that are not fullscreen would still have the configured transparency applied.

Issue #953, opened 2025-01-29
[Feature Request] Transparency


skimask9:
### Describe the problem/motivation

The new transparent effect for focused and unfocused windows is really great, but I have some questions or suggestions. Toggling transparency on some windows would be helpful, especially if it's Chrome opened with YouTube on a second monitor :) Currently, the Ignore command leaves  the window transparent.

### Describe the solution you'd like

It would be nice to ignore that window and switch transparency off or create a completely new command to toggle transparency on unfocused/focused windows

skimask9:
I have just realized that I can create a command  "set-transparency --opacity 0-100%" but it works perfectly in a focused window. I don't know how to do the same for a specific app that is unfocused :)

solemnH:
![Image](https://github.com/user-attachments/assets/559d6157-22c3-4fad-9794-23fc0c9f29c3) adding it to window_rules should do the trick

bkomadina:
> I have just realized that I can create a command "set-transparency --opacity 0-100%" but it works perfectly in a focused window. I don't know how to do the same for a specific app that is unfocused :)

could you please provide the code where you added `set-transparency --opacity 0-100%`?

I've tried the code below, move to workspace command moves Chrome to workspace 1 but the transparency doesn't change even when window_effects and focused_window transparency rules are just set to enabled without any further opacities set:
```
window_rules:
  - commands: ["move --workspace 1", "set-transparency --opacity 10%"]
    match:
      - window_process: { regex: "msedge|brave|chrome|firefox" }
```

skimask9:
> > I have just realized that I can create a command "set-transparency --opacity 0-100%" but it works perfectly in a focused window. I don't know how to do the same for a specific app that is unfocused :)
> 
> could you please provide the code where you added `set-transparency --opacity 0-100%`?
> 
> I've tried the code below, move to workspace command moves Chrome to workspace 1 but the transparency doesn't change even when window_effects and focused_window transparency rules are just set to enabled without any further opacities set:
> 
> ```
> window_rules:
>   - commands: ["move --workspace 1", "set-transparency --opacity 10%"]
>     match:
>       - window_process: { regex: "msedge|brave|chrome|firefox" }
> ```

   - commands: ["set-transparency --opacity 100%"]
    bindings: ["alt+shift+c"]

KhangHLe:
For anyone checking on this, I found a decent alternative solution building upon what the above poster said:

1. Disable transparency for focused and unfocused window

```yaml
[...]

PR #1252, opened 2026-01-21
feat: add support to ignore transparency effect on fullscreen window with new config option `ignore_fullscreen`

artinmare:
Related Feature Request: [#953 ](https://github.com/glzr-io/glazewm/issues/953) ,  [#1104 ](https://github.com/glzr-io/glazewm/issues/1104)

This PR implements a new config option `ignore_fullscreen` to allow for transparency window effect to not affect a window that is on fullscreen mode 

### Motivation
Currently, transparency can only be enabled or disabled for focused or unfocused windows regardless of whether they are in a fullscreen or not. This feature allows users to configure whether they want the transparency effect to affect a fullscreen window or not on both focused and unfocused window.

### Solution
Adds a new config option `ignore_fullscreen` that is configurable via the config file

### Example Config
Add the new `ignore_fullscreen` to the config file under the transparency options

```
window_effects:
  # Visual effects to apply to the focused window.
  focused_window:
    # Change the transparency of the window.
    transparency:
      enabled: false

      # Whether to ignore the effect on fullscreen window.
      ignore_fullscreen: false

      # Can be something like '95%' or '0.95' for slightly transparent windows.
      # '0' or '0%' is fully transparent (and, by consequence, unfocusable).
      opacity: "95%"

  # Visual effects to apply to non-focused windows.
  other_windows:
    transparency:
      enabled: false
      ignore_fullscreen: true
      opacity: "80%"
```

### Alternatives considered
Disabling transparency on both focused and unfocused windows and use window rules and keybind to achieve the transparency effect instead as shown in [#953 ](https://github.com/glzr-io/glazewm/issues/953)

### DIFF
--- a/packages/wm-common/src/parsed_config.rs
+++ b/packages/wm-common/src/parsed_config.rs
@@ -294,6 +294,9 @@ pub struct TransparencyEffectConfig {
   /// Whether to enable the effect.
   pub enabled: bool,
 
+  /// Whether to ignore the effect on fullscreen window.
+  pub ignore_fullscreen: bool,
+
   /// The opacity to apply.
   pub opacity: OpacityValue,
 }
--- a/packages/wm/src/commands/general/platform_sync.rs
+++ b/packages/wm/src/commands/general/platform_sync.rs
@@ -442,12 +442,17 @@ fn apply_transparency_effect(
   window: &WindowContainer,
   effect_config: &WindowEffectConfig,
 ) {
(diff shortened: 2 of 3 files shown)


lars-berger:
@artinmare hey, thanks for the PR!

It's a good issue to solve but I reckon we can skip adding a config option and just add this behavior by default. There's been a PR opened at #1282 for addressing the same issue - will consolidate the PRs and go ahead with closing this one

PR #1274, opened 2026-02-10
feat: workspace IPC and navigation updates

Louis047:
## Summary

Improve workspace IPC data for third-party clients and add focus commands that skip empty workspaces.

## Problem

Third-party widgets need a reliable IPC source for keep-alive workspaces and display names without parsing config. Users navigating with next/prev active workspaces also need to skip empty workspaces when keep_alive is enabled.

## Solution

- IPC: add query workspaces --include-empty to include keep_alive empty workspaces.
- IPC: enrich workspace payload with keep_alive and bind_to_monitor.
- Focus: add `focus --next-populated-workspace` and `focus --prev-populated-workspace`.
- Lifecycle: activate `keep_alive` workspaces on startup and config reload.

## Testing

- cargo build -p wm
- cargo run -p wm
- Manual: (path to run local build - `target/debug/glazewm.exe`)
  - glazewm query workspaces --include-empty
  - glazewm command focus --next-populated-workspace
  - glazewm command focus --prev-populated-workspace

## References

- YASB PR: https://github.com/amnweb/yasb/pull/679

### DIFF
--- a/packages/wm-common/src/app_command.rs
+++ b/packages/wm-common/src/app_command.rs
@@ -128,7 +128,11 @@ pub enum QueryCommand {
   /// Outputs all windows.
   Windows,
   /// Outputs all active workspaces.
-  Workspaces,
+  Workspaces {
+    /// Include empty workspaces with `keep_alive: true`.
+    #[clap(long, action)]
+    include_empty: bool,
+  },
   /// Outputs whether the window manager is paused.
   Paused,
 }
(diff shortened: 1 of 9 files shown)


Louis047:
#### Note on IPC workspace IDs 

for config‑only keep_alive workspaces (returned only when `--include-empty` is used), the [id](https://github.com/glzr-io/glazewm/tree/main/packages/wm/src/commands/container/detach_container.rs?plain=1#L34) is a deterministic hash of the workspace name. Once a workspace becomes active, it receives a new runtime UUID. IPC clients should therefore treat [name](https://github.com/glzr-io/glazewm/tree/main/packages/wm-common/src/app_command.rs?plain=1#L15) as the stable identifier (or expect ID changes when an inactive workspace becomes active).

PR #1282, opened 2026-02-21
fix: reset window effects for fullscreen, keep hide_title_bar

### LINKED ISSUE #778: [Feature Request] Conditional window effects

In Windows 11 it is possible to create a colored border around the currently focused window.

When this effect is on, it also applies to windows in fullscreen mode, such as games.

It would be nice if it was possible to either
1) disable window effects for full screen windows
2) specify conditions on when to apply window effects

I like option 1 the most, because it is simpler, and because option 2 would require the config to specify each app where window effects should not apply.
Also, I don't see the scenario where it makes sense to apply window effects to a full screen window.

### Alternatives considered

_No response_

johnsideserf:
## Summary

- For fullscreen windows, actively resets border/corner/transparency effects to defaults
- Continues to apply `hide_title_bar` for fullscreen windows
- Removed redundant ignored windows check (already excluded upstream)

## Changes

Only `packages/wm/src/commands/general/platform_sync.rs` is modified:

1. Removed `state: &WmState` parameter from `apply_window_effects()` (no longer needed)
2. Added fullscreen branch that resets border, corner style, and transparency to defaults while still applying `hide_title_bar`
3. Handles the case where effects were already applied before a window transitions to fullscreen

## Test plan

- [x] Normal tiled window has focus border as expected
- [x] Fullscreen window does NOT get border/transparency/corner effects
- [x] `hide_title_bar` still applies for fullscreen windows
- [x] Toggling back from fullscreen to tiling restores effects
- [x] Tested with Baldur's Gate 3, Cyberpunk 2077, Caves of Qud, and Satisfactory (borderless + exclusive fullscreen)

Fixes #778

### DIFF
--- a/packages/wm-platform/src/platform_impl/windows/mouse_listener.rs
+++ b/packages/wm-platform/src/platform_impl/windows/mouse_listener.rs
@@ -182,11 +182,10 @@ impl MouseListener {
     // Ignore if input is invalid or not a mouse event. Inputs from our own
     // process are also ignored, since `NativeWindow::focus` simulates
     // mouse input.
-    #[allow(clippy::cast_possible_truncation)]
     if res_size == 0
       || raw_input_size == u32::MAX
       || raw_input.header.dwType != RIM_TYPEMOUSE.0
-      || unsafe { raw_input.data.mouse.ulExtraInformation } as u32
+      || unsafe { raw_input.data.mouse.ulExtraInformation }
         == FOREGROUND_INPUT_IDENTIFIER
     {
       return Ok(());
--- a/packages/wm/src/commands/general/platform_sync.rs
+++ b/packages/wm/src/commands/general/platform_sync.rs
@@ -516,6 +516,7 @@ fn apply_window_effects(
   config: &UserConfig,
 ) {
   let window_effects = &config.value.window_effects;
+  let is_fullscreen = matches!(window.state(), WindowState::Fullscreen(_));
 
   // LINT: `effect_config` is only used on Windows.
   #[cfg_attr(not(target_os = "windows"), allow(unused_variables))]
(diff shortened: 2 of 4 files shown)


lars-berger (changes requested):

lars-berger:
@johnsideserf thank you for the PR! 🙏 

there's a similar PR open at #1252, but it'd make sense to make this the default behavior versus adding an extra config option.

left a couple of review comments to address.

johnsideserf:
Addressed review feedback:
  - Removed ignored window check (was already handled)
  - Fullscreen windows now reset active effects (border, corner, transparency) to defaults
  - hide_title_bar continues to apply for fullscreen windows
  - Effects are only reset when enabled for focused or other_windows

johnsideserf:
@lars-berger Just wanted to check in on this one - all four review comments have been addressed:

1. Removed the redundant ignored windows check
2. Fullscreen windows now actively reset border/corner/transparency effects to defaults
3. `hide_title_bar` continues to apply for fullscreen windows
4. Effects are only reset when enabled for either `focused_window` or `other_windows`

Rebuilt and tested locally today, confirmed working with fullscreen games. Would you be able to take another look when you get a chance? Thanks!

lars-berger:
@johnsideserf i appreciate the fixes, but please test this a bit more thoroughly next time. if you fullscreen or maximize a window, the effect still gets applied - it currently only applies if the window is _already_ fullscreen/maximized when focus is changed. let me know if i can help with getting you set up locally.

johnsideserf:
@lars-berger apologies, missed this during my testing, thanks for flagging.  Fixed by calling `queue_focused_effect_update()` in `update_window_state` whenever a window's state changes.  This should mean effects are re-evaluated on any state transition (tiling, floating, fullscreen, minimized), not just on focus change.

Also rebased onto latest main.

Tested locally and confirmed:

- Tiled window gets focus border normally as expected
- Fullscreening a focused window immediately removes border/transparency/corner effects
- Focus change while fullscreen does not re-apply the effects
- Toggling back from fullscreen restores effects
- Tested with the same fullscreen games noted above (borderless + exclusive)

PyricRL:
I noticed this issue as well, except whenever I use my custom binding to go into fullscreen. I currently have it set to `alt + f` and when doing this it brings the focused window into "fullscreen" but it is not maximized and therefore still has a border around it. Was wondering if there was a way to distinguish between maximizing a window and putting it in fullscreen. Thanks!

Edit:

Just realized that adding the flag `--maximized` to the command does exactly what I want. For anyone who wanted to know.

Issue #907, opened 2025-01-08
[Bug] In a two monitor setup, windows in main monitor get moved to a new workspace in secondary monitor after lock/sleep and unlock/wakeup


sodiumnitrate:
### Describe the bug

I have two monitors, and one workspace each. 1 on monitor 1 (defined as main display in windows settings), 2 on monitor 2. If I have applications open on both workspaces, and then the screens shut down either (a) because I do win->power-> lock, or (b) screen is turned off due to time limit set in system > power, upon turning the screens back on, the following happens:

- monitor 1 shows a new empty workspace called 3, which did not exist before
- monitor 2 shows 2 (active) with an option to show 1, which has the windows that were displayed on monitor 1 previously

I know this is a very specific scenario, but I'm assuming there are lots of people who daily drive more than one monitor using glazewm. I would greatly appreciate if someone could at least suggest a workaround.

### Reproduction

- Having two monitors, with monitor 1 set as main in windows settings arrange windows such that: workspace 1 on main with tiled windows & workspace 2 on secondary with tiled windows
- Win -> Power -> Lock and wait for displays to shut down
- Hit a keyboard key to wake up system and login

### Version number

3.5.0.0

xh1lr8-alt:
I have this issue as well, it is particularly egregious when it happens to programs where I need to manually save things.

bartdorsey:
This also happens to me, I have three monitors, and often the workspaces will end up on the wrong monitor after waking the machine up from sleep.

al3rez:
this is a serious issue for me

grocedj:
this has been an issue for years

jukrb0x:
I occasionally encounter this issue on my Windows 11 while using the source-built version from the main branch.

```
2026-01-07T09:58:11.606919Z  INFO glazewm::events::handle_display_settings_changed: Display settings changed.
2026-01-07T09:58:11.609216Z  INFO glazewm::commands::monitor::update_monitor: Monitor updated: Monitor(handle=1679167969, device_name=\\.\DISPLAY1, device_path=Some("\\\\?\\DISPLAY#MonitorModel#5&2486edd8&0&UID397568#{A-B-C-D-E}"), hardware_id=Some("MonitorModel"))
[...]

Issue #318, opened 2023-05-24
Unhandled exception: At least 1 workspace is required per monitor.


pter1342:
There is an issue when I am connecting my windows 10 laptop with an already running instance of glaze to a dock with two monitors, glaze crashes:

![tmp_1684925088731](https://github.com/lars-berger/GlazeWM/assets/99407909/6700fc5c-751a-400d-a2f9-572f9e1dfcb0)

After this I need to run glaze again, which works with the monitor but obviously all my windows go back to workspace 1 and I need to rearrange them again. This happens every time I disconnect the laptop from the dock and reconnect it back.

Any idea what could be the issue? here's what gets added to the log everytime a crash happens:

```
22. 5. 2023 16:17:23 
GlazeWM.Infrastructure.Exceptions.FatalUserException: At least 1 workspace is required per monitor. 
    at GlazeWM.Domain.Monitors.CommandHandlers.AddMonitorHandler.ActivateWorkspaceOnMonitor(Monitor monitor) in C:\Users\larsb\repos\GlazeWM\GlazeWM.Domain\Monitors\CommandHandlers\AddMonitorHandler.cs:line 58 
    at GlazeWM.Domain.Monitors.CommandHandlers.AddMonitorHandler.Handle(AddMonitorCommand command) in C:\Users\larsb\repos\GlazeWM\GlazeWM.Domain\Monitors\CommandHandlers\AddMonitorHandler.cs:line 45 
    at GlazeWM.Infrastructure.Bussing.Bus.Invoke[T](T command) in C:\Users\larsb\repos\GlazeWM\GlazeWM.Infrastructure\Bussing\Bus.cs:line 48 
    at GlazeWM.Domain.Monitors.CommandHandlers.RefreshMonitorStateHandler.Handle(RefreshMonitorStateCommand command) in C:\Users\larsb\repos\GlazeWM\GlazeWM.Domain\Monitors\CommandHandlers\RefreshMonitorStateHandler.cs:line 41 
    at GlazeWM.Infrastructure.Bussing.Bus.Invoke[T](T command) in C:\Users\larsb\repos\GlazeWM\GlazeWM.Infrastructure\Bussing\Bus.cs:line 48 
[...]

PR #1284, opened 2026-02-24
feat: persist workspace-to-monitor assignments across disconnections

### LINKED ISSUE #1130: [Bug] inconsistancy with connecting and disconnecting external monitors

durdn:
## Summary

When monitors temporarily disconnect (PC sleep, monitor standby, dock/undock), GlazeWM loses track of which workspaces were on which monitors. The current code treats every disconnection as permanent — it removes the monitor from the container tree and scatters its workspaces to remaining monitors. On reconnection, the monitor is treated as brand new and gets whatever workspace the config dictates, not what was actually there. This causes disruptive workspace shuffling that forces users to manually rearrange their layout after every sleep/wake cycle.

This PR introduces a **Disconnected Monitor Registry** — a lightweight record kept alongside the container tree that remembers which workspaces belonged to which monitor. When a monitor disconnects, we capture its identity (`device_path`, `hardware_id`, `device_name`) and its workspace names ordered by focus. When the same physical monitor reconnects, we restore its workspaces automatically.

### Key design decisions

- **No ghost monitors in the container tree.** The registry is a simple `Vec<DisconnectedMonitor>` on `WmState`. This avoids breaking `sort_monitors`, `monitor.index()`, `to_rect()`, and all other code that assumes every monitor in the tree is live.
- **Config bindings take priority.** `bind_to_monitor` workspace configs are processed first; ghost restoration fills in the remaining workspaces after. This means user config is always respected.
- **Matching is robust.** We match reconnecting monitors by `device_path` first (most reliable), then fall back to `hardware_id` only when it's unique across all connected + disconnected monitors.
- **Self-cleaning.** Ghost records are deduplicated by `device_path` on each disconnect and pruned after 1 hour to prevent unbounded growth.
- **`remove_monitor` is untouched.** The new `disconnect_monitor` wraps it — captures the record, then delegates to the existing removal logic.

### Edge cases handled

All monitors disconnect (sleep) — Last monitor preserved (existing logic). Others get ghost records. On wake, ghosts restore workspaces.

User moves windows while monitor is off — Ghost stores workspace names, not windows. Workspaces return with whatever windows they now contain.

Empty workspace was garbage-collected — Ghost has the name but workspace no longer exists — silently skipped.

Same monitor disconnects repeatedly — Ghost record is replaced (deduped by device_path).

Duplicate hardware IDs (e.g. identical monitors) — device_path match is preferred. hardware_id only used when unique.

Fixes #1130, #907, #318.

## Test plan

- [ ] Build: `cargo build` passes
- [ ] Lint: `cargo clippy` passes with no new warnings
- [ ] Multi-monitor: place windows on each monitor, disconnect one, verify workspaces move to remaining monitor, reconnect, verify workspaces return to the correct monitor
- [ ] Sleep/wake: put PC to sleep with multi-monitor layout, wake up, verify full layout restoration
- [ ] Single monitor: disconnect/reconnect the only monitor — no crash, layout preserved

### DIFF
--- a/packages/wm/src/commands/monitor/disconnect_monitor.rs
+++ b/packages/wm/src/commands/monitor/disconnect_monitor.rs
@@ -10,7 +10,7 @@ use crate::{
   wm_state::{DisconnectedMonitor, WmState},
 };
 
-const GHOST_TTL: Duration = Duration::from_secs(3600);
+const GHOST_TTL: Duration = Duration::from_hours(1);
 
 /// Records a monitor's identity and workspaces before removing it from
 /// the container tree. This allows workspace restoration when the same
@@ -21,7 +21,11 @@ pub fn disconnect_monitor(
   config: &UserConfig,
 ) -> anyhow::Result<()> {
   let props = monitor.native_properties();
+  #[cfg(target_os = "macos")]
+  let device_uuid = props.device_uuid.clone();
+  #[cfg(target_os = "windows")]
   let device_path = props.device_path.clone();
+  #[cfg(target_os = "windows")]
   let hardware_id = props.hardware_id.clone();
   let device_name = props.device_name.clone();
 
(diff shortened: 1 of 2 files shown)


durdn:
I will add some corollary after using this patch myself for the whole of today. I am so happy finally my PC goes in hibernation or sleep, the monitors switch off (I have 4) and when the system reactivates all my windows and workspaces are in the exact same place. I really really really am so happy.

johnazule:
Cannot wait for this, biggest pain point with glaze atm

durdn:
> Cannot wait for this, biggest pain point with glaze atm

I've been running this on my fork for a while and really love it!  

(GlazeWM/Zebar show up with some painful memory leaks/crashes every once in a while which I am slowly tracking down, I mean Codex/Claude are tracking down) but my overall experience is great. I am in heaven haha.

PR #1304, opened 2026-03-16
fix: mark window as having custom placement if `centered: false`

babysnakes:
A solution suggested to me by @lars-berger as a possible alternative to #1166. This is the part I care about the most from that PR. 

**Description:**

The current default behavior for floating windows is that if the window was not moved since launched, it will get centered occasionally because of device connect/disconnect events. If however, the user has moved the window, it has a flag set so it will not get centered. This PR sets the default `has_custom_floating_placement` flag to be true, provided that the user has configured floating centered to be false (like this):

```yaml
window_behavior:
  initial_state: "floating"
  state_defaults:
    floating:
      centered: false
```

For context see [this comment](https://github.com/glzr-io/glazewm/pull/1166#issuecomment-3703846837) from the original PR. 

Thanks

### DIFF
--- a/packages/wm/src/commands/window/manage_window.rs
+++ b/packages/wm/src/commands/window/manage_window.rs
@@ -235,7 +235,7 @@ fn create_window(
       border_delta,
       None,
       floating_placement,
-      false,
+      !prefers_centered,
       Vec::new(),
       None,
     )


babysnakes:
BTW, I have a test failing (`matches_corner_positions`) but it fails for me also in _main_ so I'm assuming it's not related to my change.

lars-berger:
awesome 🙌 thank you!

PR #1308, opened 2026-03-16
fix: catch panics in event handler to prevent process abort

durdn:
## Summary

- Wraps `process_event` in `std::panic::catch_unwind` so that panics during event handling (e.g. `RefCell` double-borrow during sleep/wake monitor transitions) are caught and logged instead of aborting the process.
- After catching a panic, the error is returned as `Err(...)` so the main loop logs it and continues — the WM recovers from a single failed event.

## Test plan

- [ ] Trigger a sleep/wake cycle with multiple monitors and verify glazewm continues running if a panic would previously have caused a crash.

### DIFF
--- a/packages/wm/src/wm.rs
+++ b/packages/wm/src/wm.rs
@@ -1,6 +1,8 @@
+use std::panic::{self, AssertUnwindSafe};
+
 use anyhow::{bail, Context};
 use tokio::sync::mpsc::{self};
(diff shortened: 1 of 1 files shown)


lars-berger:
in what scenario are you encountering a `RefCell` double-borrow? this doesn't seem correct. either way, we shouldn't be catching panics during event handling

PR #1309, opened 2026-03-16
fix: terminate IPC subscription tasks on send failure

durdn:
## Summary

- IPC subscription tasks were never terminated when a WebSocket client disconnected. `response_tx.send()` failure was wrapped as `Ok(Err(...))`, so the `if let Err` break was never reached.
- Every client reconnection (e.g. zebar) left a zombie task alive for the entire session, each receiving every `WmEvent`. Over long sessions this accumulated unbounded committed memory.
- Fix: use `.and_then()` to flatten the send error so it propagates as `Err(...)` and the task exits cleanly.

## Test plan

- [ ] Connect and disconnect an IPC subscriber (e.g. zebar) multiple times and verify memory usage stays flat.

### DIFF
--- a/packages/wm/src/ipc_server.rs
+++ b/packages/wm/src/ipc_server.rs
@@ -282,9 +282,6 @@ impl IpcServer {
                 if events.contains(&event_type)
                   || events.contains(&SubscribableEvent::All)
                 {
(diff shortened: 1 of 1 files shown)


lars-berger:
@durdn thanks for the PR! 🙏

PR #1311, opened 2026-03-16
fix: handle transient display device errors during sleep/wake

durdn:
## Summary

During sleep/wake, `EnumDisplayDevicesW` can temporarily return no active devices for a monitor that is being reinstated by the driver. Previously this caused `handle_display_settings_changed` to propagate a `DisplayDeviceNotFound` error, which triggered a blocking `MessageBoxW(MB_SYSTEMMODAL)` dialog from the tokio worker thread — freezing all user input and destabilising the WM.

- In `handle_display_settings_changed`: replace `?` on `NativeMonitorProperties::try_from()` with a `match` that logs a warning and `continue`s — the next `DisplaySettingsChanged` event re-processes the display once it is stable.
- Remove `show_error_dialog` from the non-fatal event handler error path — non-fatal errors are now logged only, with no blocking dialog.

## Test plan

- [ ] Sleep/wake with multiple monitors and verify glazewm continues running without any blocking dialog.

### DIFF
--- a/packages/wm/src/main.rs
+++ b/packages/wm/src/main.rs
@@ -281,12 +281,9 @@ async fn start_wm(
       },
     };
 
-    // Errors from event handlers are non-fatal and only logged. A blocking
-    // system modal (`show_error_dialog`) is not appropriate here because
-    // it would freeze the tokio runtime, preventing all other event
-    // processing until dismissed.
     if let Err(err) = res {
       tracing::error!("{:?}", err);
+      dispatcher.show_error_dialog("Non-fatal error", &err.to_string());
     }
   }
 
--- a/packages/wm/src/events/handle_display_settings_changed.rs
+++ b/packages/wm/src/events/handle_display_settings_changed.rs
@@ -18,27 +18,27 @@ pub fn handle_display_settings_changed(
 ) -> anyhow::Result<()> {
   tracing::info!("Display settings changed.");
 
-  // Ignore the event if retrieval of the displays fails.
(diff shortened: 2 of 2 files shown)


packages/wm/src/main.rs
```diff
@@ -283,7 +283,6 @@ async fn start_wm(
 
     if let Err(err) = res {
       tracing::error!("{:?}", err);
-      dispatcher.show_error_dialog("Non-fatal error", &err.to_string());
```
lars-berger: this doesn't make any sense unfortunately

freddiehaddad:
In simple terms what's happening is if your computer is waking from sleep or the monitor is waking from low power state there's a race where the monitor isn't ready, yet when the code is executing. In this case, the monitor isn't found a dialogue error message appears. This PR is sending that message to a log instead, so you don't see a UI popup for a transient error.

One question is that this started happening in 3.10. Did something change that suddenly causes this issue to surface? Why did it not happen in earlier versions?

freddiehaddad:
For record, this is the issue. When I unlocked my computer and signed in (display was asleep):
image

durdn:
> For record, this is the issue. When I unlocked my computer and signed in (display was asleep): image

exactly that for me as well.

jondmarien:
> For record, this is the issue. When I unlocked my computer and signed in (display was asleep):
> image
> 

I know this was pushed and fixed in the latest update, but 3.1 did NOT introduce this for me. I have been dealing with this bug for over a year. Happy someone put up a PR for it lol.

PR #1318, opened 2026-03-17
fix: restore `lwin`/`rwin`/`win` keybindings (regression) (#1315)

### LINKED ISSUE #1315: [Bug] win-keys broken on 3.10.0
### Describe the bug

i'm just updated from 3.9.1 to 3.10.0 and all my win-keys doesn't work
rolling back to 3.9.1 solve problem

### Reproduction

just update to 3.10.0

example binding:

```
keybindings:
  # Close focused window.
  - commands: ["close"]
    bindings: ["lwin+q", "rwin+q"]
```

```shell

```

### Version number

3.10.0

freddiehaddad:
Make `win`, `lwin`, and `rwin` aliases for `Cmd`, `LCmd`, and `RCmd` respectively, since they map to the same physical key on all platforms.

Previously, these were separate `Key` enum variants that shared the same key codes as their `Cmd` counterparts. This caused keybindings using `lwin`/`rwin` to fail because the keyboard hook resolved `VK_LWIN` to `Key::Cmd` (never `Key::LWin`), and the modifier rejection logic treated `LCmd` as an unwanted modifier since it wasn't in the binding.

Fixes #1315

### DIFF
--- a/packages/wm-platform/src/models/key.rs
+++ b/packages/wm-platform/src/models/key.rs
@@ -118,6 +118,7 @@ pub enum Key {
   Ctrl,
   Alt,
   Shift,
+  Win,
   LCmd,
   RCmd,
   LCtrl,
@@ -126,6 +127,8 @@ pub enum Key {
   RAlt,
   LShift,
   RShift,
+  LWin,
+  RWin,
 
   // Special keys
   Space,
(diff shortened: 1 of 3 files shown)


lars-berger:
@freddiehaddad thanks for the help with tracking this down 🙏. there's a simpler fix that preserves the `Display` formatting, so I've added some touchups

PR #1320, opened 2026-03-17
fix: update objc2 dependencies for macOS 26 compatibility

lohrm-spf:
## Summary

- Bumps `objc2` from 0.6.1 to 0.6.4 and all `objc2-*` framework crates from 0.3.1 to 0.3.2
- Removes unnecessary `unsafe` blocks that are no longer required in the updated API

## Problem

GlazeWM crashes on startup on macOS 26 (Tahoe) with:

```
invalid message send to -[_TtGCs23_ContiguousArrayStorageCSo8NSScreen_$
countByEnumeratingWithState:objects:count:]: expected return to have type
code 'q', but found 'Q'
```

Apple changed the return type of `countByEnumeratingWithState:objects:count:` from `NSInteger` (signed, type code `q`) to `NSUInteger` (unsigned, type code `Q`) in macOS 26. The objc2 runtime's message send verification catches this mismatch and panics.

This is a [known issue](https://github.com/madsmtm/objc2/issues/765) fixed in objc2 0.6.2+.

## Testing

Tested on macOS 26.3.1 (MacBook Pro M-series). GlazeWM starts and enumerates monitors successfully after the update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1516,18 +1516,18 @@ dependencies = [
 
 [[package]]
 name = "objc2"
-version = "0.6.1"
+version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "88c6597e14493ab2e44ce58f2fdecf095a51f12ca57bec060a11c57332520551"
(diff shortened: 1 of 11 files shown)


lars-berger:
@michidk awesome, thank you! 🙌

PR #1347, opened 2026-04-02
fix: prevent crash on wake/startup by skipping cleanup_invalid_windows under commit pressure

durdn:
## Context

This fix is the result of several weeks of crash investigation on Windows. We tracked crash reports via Windows Error Reporting, collected minidumps, built a custom PDB RVA decoder to map fault offsets to source lines, and added unbuffered memory diagnostics (`GlobalMemoryStatusEx` fields written directly to `errors.log` every 5 seconds) to correlate system commit state with crash timing.

All crashes shared the same signature: exception `0xc0000409` (FAST_FAIL_FATAL_APP_EXIT), no Rust panic marker, no `catch_unwind` coverage. Heap dump analysis confirmed `rust_oom → handle_alloc_error → __fastfail(7)` as the call chain. The fix was validated over multiple days of sleep/wake cycles and system restarts without a single recurrence.

## Problem

On Windows, waking from sleep causes a transient virtual-memory commit pressure spike as the OS pages suspended processes back in. During this window, `ullAvailPageFile` drops sharply. Any heap allocation during peak pressure triggers `handle_alloc_error → rust_oom → __fastfail(7)` (exception `0xc0000409`), terminating the process. This cannot be caught by `catch_unwind`.

The cleanup tick runs `cleanup_invalid_windows` every 5 seconds. This iterates the container tree and allocates during window state management — if it fires during a pressure spike, it OOMs and crashes glazewm on every startup or wake.

Root cause was confirmed via Windows Error Reporting minidumps and a custom PDB RVA decoder. The `spawn_pipe_relay +0x2291` label seen in earlier crash reports is a **PDB nearest-symbol artifact** — `rust_oom → intrinsics::abort()` compiles to `int 0x29` at that binary address. No relay thread was ever present.

## Solution

Add `Dispatcher::is_under_commit_pressure()` to `wm-platform`, which calls `GlobalMemoryStatusEx` and returns `true` when `ullAvailPageFile < ullTotalPageFile / 2`. Returns `false` unconditionally on macOS.

When commit pressure is high, the cleanup tick skips `cleanup_invalid_windows` entirely. The work is **deferred, not dropped** — the next tick retries 5 seconds later once pressure has subsided.

## Testing

- Confirmed crash-free across dozens of sleep/wake cycles and system restarts over multiple days on a Windows 11 machine.
- `is_under_commit_pressure()` returns `false` on macOS (no-op on that platform).

## Notes

- A companion PR (#1348) adds a debounce on `DisplaySettingsChanged` using the same `is_under_commit_pressure()` API, as the handler itself can also OOM during pressure spikes.
- Also fixes a pre-existing `clippy::cast_possible_truncation` lint in `mouse_listener.rs` (`ulExtraInformation` is already `u32`).

### DIFF
--- a/packages/wm-platform/Cargo.toml
+++ b/packages/wm-platform/Cargo.toml
@@ -32,6 +32,7 @@ windows = { version = "0.52", features = [
   "Win32_Graphics_Gdi",
   "Win32_Security",
   "Win32_System_Com",
+  "Win32_System_SystemInformation",
   "Win32_System_Environment",
   "Win32_System_LibraryLoader",
   "Win32_System_Registry",
--- a/packages/wm-platform/src/dispatcher.rs
+++ b/packages/wm-platform/src/dispatcher.rs
@@ -26,7 +26,10 @@ use windows::{
   core::PCWSTR,
   Win32::{
     Foundation::POINT,
-    System::Environment::ExpandEnvironmentStringsW,
+    System::{
+      Environment::ExpandEnvironmentStringsW,
+      SystemInformation::{GlobalMemoryStatusEx, MEMORYSTATUSEX},
+    },
     UI::{
       Input::KeyboardAndMouse::{
         GetAsyncKeyState, VK_LBUTTON, VK_RBUTTON,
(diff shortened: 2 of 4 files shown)


durdn:
Before this (and the sister fix) my GlazeWM would crash almost every time my PC went through a sleep/wake cycle. On my local fork with this applied it hasn't crashed since. I think Claude finally found the root cause, it went through dozens of crashes during several weeks with a bunch of low level techniques to find out the root cause.

lars-berger:
is this reproducible with the changes from #1312? would imagine the interval tick buildup is largely the reason why this could cause issues

durdn:
> is this reproducible with the changes from #1312? would imagine the interval tick buildup is largely the reason why this could cause issues

yes the sleep/wake crashes kept happening even with #1312 present in my fork,

lars-berger:
hmm interesting. i'm honestly hesitant to merge these (#1347 and #1348), since they're quite large changes and i haven't received reports on Discord or in the GH issues from others running into this. could it be something environment-specific that's interfering with this?

i'll go ahead and merge the #1312 PR though 👍

durdn:
> hmm interesting. i'm honestly hesitant to merge these (#1347 and #1348), since they're quite large changes and i haven't received reports on Discord or in the GH issues from others running into this. could it be something environment-specific that's interfering with this?
> 
> i'll go ahead and merge the #1312 PR though 👍

Fair. If new evidence/crash reports come up you can always consider them at a later stage 👌

lars-berger:
true, i'll leave them open for now then 👍

PR #1367, opened 2026-04-07
test: adding bon-based mocks for model types

popiel:
This adds bon-based mocks for the model types (containers and their properties).  It also fixes a few doc test failures and clippy artifacts.

### DIFF
--- a/packages/wm/Cargo.toml
+++ b/packages/wm/Cargo.toml
@@ -42,3 +42,7 @@ wm-common = { path = "../wm-common" }
 wm-ipc-client = { path = "../wm-ipc-client" }
 wm-macros = { workspace = true }
 wm-platform = { path = "../wm-platform" }
+
+[dev-dependencies]
+bon = "3"
+wm-platform = { path = "../wm-platform", features = ["test_utils"] }
--- a/packages/wm/src/test_utils.rs
+++ b/packages/wm/src/test_utils.rs
@@ -1,0 +1,260 @@
+//! Test utilities for creating mock container instances.
+//!
+//! This module provides default values and helper functions used by the
+//! mock builders in the model modules.
+
+use bon::bon;
+use wm_common::{
+  FloatingStateConfig, GapsConfig, TilingDirection, WindowState,
+  WorkspaceConfig,
+};
+use wm_platform::{Display, NativeWindow, Rect, RectDelta};
+
+use crate::{
+  commands::container::attach_container,
+  models::{
+    Monitor, NativeMonitorProperties, NativeWindowProperties,
+    NonTilingWindow, SplitContainer, TilingContainer, TilingWindow,
+    Workspace,
+  },
+  traits::TilingSizeGetters,
+};
+
+pub const MOCK_MONITOR_WIDTH: i32 = 1680;
+pub const MOCK_MONITOR_HEIGHT: i32 = 1050;
+pub const MOCK_TASKBAR_HEIGHT: i32 = 50;
+pub const MOCK_DPI: u32 = 96;
+pub const MOCK_SCALE_FACTOR: f32 = 1.0;
+pub const MOCK_WINDOW_WIDTH: i32 = 300;
+pub const MOCK_WINDOW_HEIGHT: i32 = 200;
+
+pub fn mock_bounds() -> Rect {
+  Rect::from_xy(0, 0, MOCK_MONITOR_WIDTH, MOCK_MONITOR_HEIGHT)
+}
+
+pub fn mock_working_area() -> Rect {
+  Rect::from_xy(
+    0,
+    0,
+    MOCK_MONITOR_WIDTH,
+    MOCK_MONITOR_HEIGHT - MOCK_TASKBAR_HEIGHT,
+  )
+}
+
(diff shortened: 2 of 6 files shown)


packages/wm/Cargo.toml
```diff
@@ -7,6 +7,9 @@ license = "GPL-3"
 edition = "2021"
 default-run = "glazewm"
 
+[lib]
+path = "src/lib.rs"
```
lars-berger: is it needed that we turn this into a lib?

packages/wm/src/wm.rs
```diff
@@ -780,6 +780,7 @@ impl WindowManager {
   }
 
   /// Runs cleanup tasks when the WM is exiting.
+  #[allow(dead_code)]
```
lars-berger: `cleanup` is still in use afaik

packages/wm/src/test_utils.rs
```diff
@@ -0,0 +1,299 @@
...
+      let tiling_size =
+        if distribute_tiling_sizes && !tiling_containers.is_empty() {
+          1.0 / tiling_containers.len() as f32
+        } else {
+          1.0
+        };
+
+      for child in tiling_containers {
+        let container: Container = child.into();
+        if distribute_tiling_sizes {
```
lars-berger: Tiling size is already distributed in `attach_container`

packages/wm/src/test_utils.rs
```diff
@@ -0,0 +1,299 @@
+//! Test utilities for creating mock container instances.
+//!
+//! This module provides default values and helper functions used by the
+//! mock builders in the model modules.
+
+#[cfg(test)]
+#[allow(dead_code)]
+pub(crate) mod mocks {
```
lars-berger: reckon we can avoid the wrapper mods

packages/wm/src/test_utils.rs
```diff
@@ -0,0 +1,299 @@
...
+  use wm_platform::{Rect, RectDelta};
+
+  /// Default monitor dimensions (1680x1050).
```
lars-berger: avoid comments that simply re-state the value of the const

packages/wm/src/test_utils.rs
```diff
@@ -0,0 +1,299 @@
...
+  pub const DEFAULT_MONITOR_WIDTH: i32 = 1680;
+  pub const DEFAULT_MONITOR_HEIGHT: i32 = 1050;
+
+  /// Taskbar height offset for working area.
+  pub const TASKBAR_HEIGHT: i32 = 50;
```
lars-berger: think it’d be fine to get rid of this. if we really need to simulate the taskbar being present, we can do it within the test itself

packages/wm/src/test_utils.rs
```diff
@@ -0,0 +1,299 @@
...
+
+  /// Default DPI for monitors.
+  pub const DEFAULT_DPI: u32 = 96;
```
lars-berger: do these consts need to be `pub`?

packages/wm/src/test_utils.rs
```diff
@@ -0,0 +1,299 @@
...
+
+  /// Default scale factor.
+  pub const DEFAULT_SCALE_FACTOR: f32 = 1.0;
```
lars-berger: reckon we should rename `DEFAULT_xxx` to some other prefix to make it clearer that these are arbitrary values. maybe just `MOCK_xxx`?

packages/wm/Cargo.toml
popiel: It was when I had stuff a little more spread around.  Trying again to eliminate it.

packages/wm/src/wm.rs
popiel: That was an artifact of adding lib stuff.  Now that that's gone, this is unneeded.

packages/wm/src/test_utils.rs
popiel: Nuked.

packages/wm/src/test_utils.rs
popiel: Nuked.

packages/wm/src/test_utils.rs
popiel: In my experience, many bugs arise precisely because most tests neglect details like the possibility that the working area is smaller than the full size.

packages/wm/src/test_utils.rs
popiel: If a test needs to reference them, I'd rather not have added friction.

packages/wm/src/test_utils.rs
popiel: Renamed.

packages/wm/src/test_utils.rs
popiel: Working on it.

packages/wm/Cargo.toml
popiel: Gone.

packages/wm/src/test_utils.rs
popiel: Unwrapped.

PR #1393, opened 2026-05-19
Restore windows to their original state on GlazeWM exit

Hsc717:
fix: Restore windows to their original state on GlazeWM exit

Summary

This PR fixes a long-standing issue where windows are left in their tiled positions (and lose their maximized state) when GlazeWM exits. Previously, on shutdown, windows were redrawn at their WM-managed layout positions instead of being restored to where they were before GlazeWM started.

Problem

When GlazeWM exits, all managed windows remain in the positions assigned by the tiling layout. Maximized/fullscreen windows permanently lose their maximized state — they are left as non-maximized windows at their tiled size/position, disrupting the user's desktop layout.

Changes

1. Snapshot original window state at startup

**`packages/wm/src/models/native_window_properties.rs`**:
- Added `original_frame: Rect` — captures the window's frame position and size before GlazeWM takes over management. This is populated once during initial state population via `TryFrom<&NativeWindow>`.
- Added `original_is_maximized: bool` — records whether the window was originally maximized before being converted to tiling mode.

 2. Restore windows on exit

**`packages/wm/src/wm_state.rs`** — `Drop for WmState`:
- Instead of redrawing windows at their tiled positions (`window.to_rect()`), the `Drop` implementation now restores each window using its cached `original_frame`.
- **On Windows**: Windows that were originally maximized are restored using `restore(Some(&original_frame))` to set the normal bounds, followed by `maximize()` to restore the maximized state. Non-maximized windows use `set_frame(&original_frame)`.
- Effects (borders, transparency, taskbar visibility) are still reset as before.
- This also unhides windows that were on other workspaces (macOS).

## Testing

- Verified that windows return to their exact original position and size after `exit`.
- Confirmed that maximized windows are properly restored to maximized state (not left as tiled non-maximized windows).
- Builds cleanly with no new dependencies.

### DIFF
--- a/packages/wm/src/models/native_window_properties.rs
+++ b/packages/wm/src/models/native_window_properties.rs
@@ -9,8 +9,14 @@ pub struct NativeWindowProperties {
   pub class_name: String,
   pub process_name: String,
   pub frame: Rect,
+  /// The window's original position/size before GlazeWM managed it.
+  /// Used to restore windows to their initial state when the WM exits.
+  pub original_frame: Rect,
   pub is_minimized: bool,
   pub is_maximized: bool,
+  /// Whether the window was originally maximized before GlazeWM managed it.
+  /// Used to restore maximized state when the WM exits.
+  pub original_is_maximized: bool,
   pub is_resizable: bool,
   #[cfg(target_os = "windows")]
   pub shadow_borders: RectDelta,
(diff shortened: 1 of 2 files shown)


Otamono:
can this be disabled via the config? Cause i rather actually have the windows stay in their tiled position when exiting. Cause currently its broken and it slightly resizes and moves the windows. But definitly not would want them to restore to the chaos they were in before starting glaze wm.

PR #1396, opened 2026-05-22
fix: make system tray icon monochrome

### LINKED ISSUE #1395: [macOS] System tray / menu bar icon is colored instead of monochrome (template icon)
### Description
On macOS, the GlazeWM menu bar icon is currently displayed as a full-color image (the default orange/purple gradient logo). This doesn't align with the native macOS design guidelines, where status menu items (menu bar extras) are expected to be monochrome and adapt dynamically to light/dark system themes.

### Root Cause
In `packages/wm/src/sys_tray.rs`, the tray icon is instantiated via the `tray-icon` crate using `TrayIconBuilder`. However, it does not instruct the library to treat the icon as a template on macOS. 

On macOS, when an icon is configured as a template, the OS automatically ignores its original colors and uses the alpha channel to render it in black (light mode) or white (dark mode) depending on the user's system preferences.

### Proposed Solution
We can fix this by conditionally configuring `with_icon_as_template(true)` for macOS targets when building the tray icon.

*(Note: We must use the `#[cfg(target_os = "macos")]` block rather than `cfg!(target_os = "macos")` because the `with_icon_as_template` method is only compiled and available on macOS in the `tray-icon` crate API, and would otherwise cause compilation failures on Windows/Linux targets).*

In `packages/wm/src/sys_tray.rs`:

```rust
    let mut builder = TrayIconBuilder::new()
      .with_menu(Box::new(tray_menu))
      .with_tooltip(format!("GlazeWM v{}", env!("VERSION_NUMBER")))
      .with_icon(icon);

    #[cfg(target_os = "macos")]
    {
      builder = builder.with_icon_as_template(true);
    }

    let tray_icon = builder.build()?;
```

mikejobrn:
### Summary
This PR configures the system tray icon to render as a template on macOS, making it monochrome and allowing it to adapt dynamically to macOS light/dark themes.

### Details
- Configures `TrayIconBuilder::with_icon_as_template(true)` using conditional compilation (`#[cfg(target_os = "macos")]`).
- Using conditional compilation is necessary since `with_icon_as_template` is only defined on the macOS target in the `tray-icon` crate; calling it directly on Windows targets would result in compilation failures.

Closes #1395

### DIFF
--- a/packages/wm/src/sys_tray.rs
+++ b/packages/wm/src/sys_tray.rs
@@ -186,11 +186,15 @@ impl SystemTray {
       "../../../resources/assets/icon.png"
     ))?;
 
-    let tray_icon = TrayIconBuilder::new()
+    let builder = TrayIconBuilder::new()
       .with_menu(Box::new(tray_menu))
       .with_tooltip(format!("GlazeWM v{}", env!("VERSION_NUMBER")))
-      .with_icon(icon)
-      .build()?;
+      .with_icon(icon);
+
+    #[cfg(target_os = "macos")]
+    let builder = builder.with_icon_as_template(true);
+
+    let tray_icon = builder.build()?;
 
     Ok(tray_icon)
   }


lars-berger:
nice! would you mind sharing a screenshot of how it looks (light & dark)?

mikejobrn:
> nice! would you mind sharing a screenshot of how it looks (light & dark)?
Sure!
Captura de Tela 2026-05-23 às 18 18 29
Captura de Tela 2026-05-23 às 17 14 31

PR #1400, opened 2026-06-11
feat: add mode to keep workspaces on the primary monitor only

hairbui76:
## Summary

Adds an opt-in mode where workspaces exist only on a single "primary" monitor, while all other monitors are left to the OS. This is useful for setups where you want GlazeWM to tile your main display (e.g. the laptop panel) but leave external monitors untouched.

## New config options

```yaml
general:
  # Whether workspaces can span multiple monitors.
  # - 'true': Each monitor manages its own workspaces (default, current behavior).
  # - 'false': Only the primary monitor receives workspaces. Windows on other
  #   monitors are left unmanaged (OS-controlled) and get managed automatically
  #   once moved onto the primary monitor.
  multi_monitor_workspaces: true

  # Optional stable identifier for the primary monitor. When unset, the
  # leftmost monitor (index 0) is used. Run `glazewm query monitors` and copy
  # `hardwareId` (Windows: EDID-derived string; macOS: CoreGraphics display UUID).
  # In my case, my main monitor hardwareId is `LEN403A`
  primary_monitor_hardware_id: 'LEN403A'
```

Defaults preserve the existing behavior - nothing changes unless
`multi_monitor_workspaces: false` is set.

## Behavior with `multi_monitor_workspaces: false`

- Workspace placement: all workspaces (including `bind_to_monitor` and `keep_alive` ones) live on the primary monitor. Secondary monitors hold no workspaces. move-workspace-in-direction between monitors is disabled.
- Windows on secondary monitors are not managed by the WM, so workspace switches on the primary monitor never affect them:
  - Dragging a managed window onto a secondary monitor releases it to the OS, snapping it to workspace-like bounds (respects outer gaps and working area).
  - Moving it back onto the primary monitor (drag release, or programmatic moves like `Win+Shift+Arrow`) re-manages it automatically.
  - Windows already on secondary monitors at startup, or newly shown there, are likewise left OS-managed and registered for re-management.
- Display topology changes apply immediately: on every display settings change (monitor plugged/unplugged, sleep/wake), workspace placement is re-enforced. E.g. if the configured primary monitor is plugged in after startup, all workspaces move onto it right away.
- Config reload: toggling the option at runtime consolidates or redistributes workspaces accordingly.

## Other changes

- New `manage` command (inverse of ignore): stops ignoring the OS foreground window and re-attaches it to the layout. Sample keybinding: `alt+shift+n`.
- `glazewm query monitors` now exposes `hardwareId` for identifying displays.

## Testing

- `cargo test --workspace` and `cargo clippy --workspace` (pedantic) pass.
- Manually verified on Windows 11 with a two-monitor setup: startup with both monitors, hot-plugging the primary/secondary in both orders, sleep/wake, dragging windows across monitors, `Win+Shift+Arrow` moves, workspace switching on the primary, and toggling the option via config reload.

### DIFF
--- a/packages/wm-common/src/app_command.rs
+++ b/packages/wm-common/src/app_command.rs
@@ -159,6 +159,11 @@ pub enum InvokeCommand {
   Close,
   Focus(InvokeFocusCommand),
   Ignore,
+  /// Stops ignoring the foreground window and attaches it to the layout.
+  ///
+  /// Uses the OS foreground window, not `GlazeWM`'s focused container,
+  /// because ignored windows are not in the WM tree.
+  Manage,
   Move(InvokeMoveCommand),
   MoveWorkspace {
     #[clap(long)]
--- a/packages/wm-common/src/dtos/monitor_dto.rs
+++ b/packages/wm-common/src/dtos/monitor_dto.rs
@@ -26,4 +26,7 @@ pub struct MonitorDto {
   pub device_path: Option<String>,
   pub hardware_id: Option<String>,
   pub working_rect: Rect,
+  /// Whether this monitor is the configured primary monitor.
+  #[serde(default)]
+  pub is_primary: bool,
 }
--- a/packages/wm-common/src/dtos/monitor_dto.rs
+++ b/packages/wm-common/src/dtos/monitor_dto.rs
@@ -24,6 +24,7 @@ pub struct MonitorDto {
   pub handle: Option<isize>,
   pub device_name: String,
   pub device_path: Option<String>,
+  /// Windows: EDID-derived hardware ID. macOS: CoreGraphics display UUID (same value used for `general.primary_monitor_hardware_id`).
   pub hardware_id: Option<String>,
   pub working_rect: Rect,
   /// Whether this monitor is the configured primary monitor.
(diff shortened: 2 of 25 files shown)


hairbui76:
Maybe this will be feature for #753

hairbui76:
### Updated: Snap newly opened windows on workspaceless monitors to tiling extent

### Detailed:

When `multi_monitor_workspaces` is disabled, only the primary monitor has a displayed workspace; windows on other monitors stay OS-managed and are queued for remanage. Previously the move/unmanage path snapped such windows
to the monitor's single-window workspace extent, but the fresh-open path in `manage_window` returned early without snapping. As a result, a window opened directly on a non-primary monitor kept its app-defined size instead of filling the screen like a tiled window.

Wire the snap into the fresh-open branch so all entry points behave consistently:

- Add `snap_new_native_window_to_external_monitor_workspace`, a variant of   the existing snap helper for windows that have no `WindowContainer` yet. It resizes the raw `NativeWindow` to `Monitor::max_workspace_rect_from_gaps` (outer gaps + working area), matching a single tiled window.
- Call it in `manage_window` right before registering the window for remanage, for windows on a monitor without a displayed workspace.
- Compensate for shadow borders (Windows only, mirroring `total_border_delta` for a zero-border window). Without this, the OS would place the shadow-inclusive frame inside the tiling rect, leaving the visible window slightly smaller than the workspace extent.

No existing behavior changes: the managed, multi-monitor, and primary-monitor paths are untouched; only the previously-unhandled fresh-open-on-external-monitor case now snaps.

Issue #678, opened 2024-08-16
[Feature Request] Layout support


lars-berger:
### Describe the problem/motivation

Windows are currently arranged via move commands and by changing the tiling direction. It would be nice to have preset layouts that windows can be automatically arranged into.

### Describe the solution you'd like

**Depends on:**
* Add `swap --direction <DIRECTION>` command - #677
* Automatically enable binding modes - #675

### Proposed solution

A workspace can optionally have a layout active, which will do 3 things:

  1) When activated (e.g. via a keybinding command) it will **automatically arrange** your windows in the current workspace to the specified layout (e.g. a grid layout).
  2) When a **new window is opened** within that workspace or moved to that workspace, it will be arranged according to the active layout.
  3) Windows can still be moved as per usual WM behavior when the layout is active, but if a window is moved in a way that doesn't correspond to the active layout (e.g. if a grid layout is active and the windows aren't in a grid anymore), then **the layout will deactivate** and windows won't be added in a grid anymore.

A crucial part to making this idea work is to be able to move windows without breaking the layout. But if there was a way to **swap** windows in a given direction (#677), the layout could be maintained.

So what would make a lot of sense IMO is to add a command for swapping windows, and to also have a binding mode that automatically triggers (#675) when a layout is active. When this binding mode is active, move keybindings are changed to swap keybindings.

**Potential config:**

```yaml
binding_modes:
  name: 'swap'
  match_enabled:
    # Automatically trigger binding mode when the focused workspace has the given layout
    - focused_workspace_layout: { exact: 'dwindle' }
  keybindings:
      # Replace alt+shift+hjkl move commands with swaps
      - command: 'swap --direction left'
        bindings: ['alt+shift+h', 'alt+shift+left']
      - command: 'swap --direction right'
        bindings: ['alt+shift+l', 'alt+shift+right']
      - command: 'swap --direction up'
        bindings: ['alt+shift+k', 'alt+shift+up']
      - command: 'swap --direction down'
        bindings: ['alt+shift+j', 'alt+shift+down']
        
keybindings:
  # Activate dwindle layout
  - commands: ['layout --name dwindle']
    bindings: [..]
```

**Potential layouts:**
* Dwindle
* Grid
* Monocle

Foo-x:
Maybe Columns/Rows layouts are useful for wide/vertical monitors.

```
Columns
┌──┐┌──┐┌──┐┌──┐
│  ││  ││  ││  │
│  ││  ││  ││  │
[...]

Issue #301, opened 2023-04-17
Stacking like in i3


danielvan1:
Is there any plans of creating the stacking abilities like in i3?

not-ski:
Huge +1 to this feature request; stacking will make Glaze _much_ more versatile!

danfbfern:
Is there a way to bring more visibility to this or have an update on it?

flaschenpost:
+1 from me.

guidokl:
+1 Stacking would be handy in many situations for me.

exic:
Any plans to implement stacking? Since I read that it's available on Mac now, I'm considering to try it but with aerospace having stacking and glazewm apparently still not, I'll have to stick to aerospace..

chrnz008:
hope this feature comes soon

Issue #233, opened 2023-01-31
Tabs


parsiad:
Are there any plans to support tabs similar to the tabbed [layout in i3](https://i3wm.org/docs/userguide.html#_changing_the_container_layout)?

notblam:
This is asked pretty often, and I agree something like this would be great.  Occasionally there is some talk about how to support things like this in the form of scriptable layouts or fully automatic layouts.  No one has implemented anything or identified what the best approach would be.  The changes needed would likely be pretty big.

asmodeus812:
That would indeed be fantastic, something that is missing by default in i3 for tabs is being able to pick a tab for example (similarly to how in browsers we can use ctrl + tab number to open / focus the tab)

gespau:
I would also love a tabbed layout. maybe it will come after #604 
If this feature were implemented, I wouldn't even need a maximize feature ...

rmanthorpe:
I would also love to see tabs. It's a feature I make use of a lot in i3wm.

khusmann:
Agreed, this is the only feature in GlazeWM missing for me! For now I'm getting a sort of pseudo-tabbed behavior (for single-screen apps) by fullscreening them, and then [alt-tabbing between them](https://github.com/glzr-io/glazewm/issues/738).

The downside is that fullscreened apps cover the zebar...

christopheleroux:
Hello, 

I'm an unconsented windows user and GlazeWM is my sunlight in the dark =)
I3wm tabbed layout like in GlazeWM would be the 🍒 on the cake .

Thank by advance if you have time to consider this issue (and maybe I would have enough time to learn rust ... in the future )

estavrionto:
need tabbed layout in glazewm pleaase

PR #1410, opened 2026-07-15
feat: add accordion tiling layout

### LINKED ISSUE #1005: [Feature Request] Hope stacking or tabbed layout will add to glazewm

Some applications GUI are just too heavy, they have to take almost entire whole screen to work comfortably. For example: Unity Editor, Visual Studio, Office suite, Adobe suite, etc.

Image

Sometimes you have to open multiple instances of these heavy GUI applications, at this moment, the number of workspaces are just not enough, and too difficult to organize.
My approach is put these multiple instances to the same workspace, minimize all of them except the one I want to work on. but switch between these instances still relies on mouse interactions with the taskbar.

The stacking/tabbed layout of i3 and the accordion layout of [aerospace](https://nikitabobko.github.io/AeroSpace/guide#layouts) are both very good for this kind of situation. 

We just need to focus each window to make sure it's maximized and on top of all others, and can look at the gap to know if there is any window behind the current one.

### Alternatives considered

There are definitely some workaround, but it takes more effort to make this happen.

kwtop:
## Summary

- Add independent `tiles` and `accordion` layout modes to every workspace and split container.
- Implement AeroSpace-style overlapping geometry in both directions, with configurable edge padding and MRU-aware window ordering.
- Expose the layout through commands, queries, subscriptions, IPC DTOs, the sample config, and documentation.
- Preserve the existing container tree and tiling ratios when switching layouts, including nested resize, move, flatten, focus, and drag/drop behavior.

## Why

Large GUI applications often need nearly the full workspace, but users still need a fast way to group and switch between several instances. Accordion layout keeps the focused window large while revealing the edges of neighboring windows, so the group remains discoverable without requiring extra workspaces or taskbar interaction.

Fixes #1005.

Related to #301, #233, and #678.

## User-facing API

- `set-tiling-layout tiles|accordion`
- `toggle-tiling-layout`
- `query tiling-layout`
- `sub -e tiling_layout_changed`
- `gaps.accordion_padding` (default: `30px`)
- `alt+b` in the sample/default config

A horizontal accordion behaves like a tabbed layout; a vertical accordion behaves like a stacked layout. Layout is an independent dimension from `TilingDirection`, so switching back to `tiles` restores the same tree and tiling proportions.

This is orthogonal to #1406: accordion changes how the existing children of a direction container are displayed, rather than declaratively rearranging the workspace tree.

## Implementation notes

- Existing configurations and older DTO payloads default to `tiles`.
- Geometry matches AeroSpace's first/last/MRU-adjacent inset behavior and clamps padding so windows retain a positive size.
- Percentage padding, DPI scaling, nested accordion resize, and padding clamp transitions are supported.
- Drag hit testing follows the topmost visible leaf window, including passthrough across inner-gap holes.
- Windows maintains full MRU subtree z-order. On macOS, only the focused window can be raised reliably through public APIs, so nested accordion subtrees may visually interleave; this limitation is documented.

## Validation

- `cargo fmt --all -- --check`
- `cargo check --workspace --all-targets --all-features`
- `cargo test --workspace --all-features`
  - 23 `wm` tests
  - 5 `wm-common` tests
  - 11 platform integration tests
  - all applicable doctests
- `cargo clippy --all-targets --all-features -- -D warnings`
- `git diff --check`


sim590:
@kwtop: can we know why this was closed and not reviewed then merged?

LPlusBonk:
@kwtop I'm curious as well, could this please be re-opened? It's a feature I am finding I could really use in my personal workflows, and personally I think you present it in an elegant and appropriate manner.

From what I can see, it's a complete solution to the relatively common request of i3-style tabbed or stacked containers.

Was it closed in anticipation of #1406 breaking it by any chance? Was it closed because it's a pretty deep change and should have some other system added first or something? Was it closed on accident?

PR #1437, opened 2026-09-20
feat: allow one instance per logged-in Windows user

herlondf:
## What this does

GlazeWM refuses to start when another Windows user is already logged in and running it. This lets each logged-in user run their own instance.

## Why it fails today

Two things are machine-wide rather than per-session.

The single instance mutex lives in the `Global\` namespace. Whoever starts GlazeWM first owns that object, and its default DACL grants no access to anyone else, so `CreateMutexW` in a second session returns `ERROR_ACCESS_DENIED`. That happens before the `ERROR_ALREADY_EXISTS` check, which is why the log shows a raw Windows error rather than the friendly message:

```
ERROR glazewm: Acesso negado. (0x80070005)
```

The IPC server also binds a fixed port, and a port belongs to the machine and not the session. The second instance cannot bind, and the error propagates out of `start_wm` taking the whole WM with it. Fixing only the mutex is not enough.

## What changed

The mutex moved to the `Local\` namespace, which is per-session. Running two instances in the same session is still blocked, exactly as before.

The IPC server now keeps port 6123 whenever it is free and falls back to an OS-assigned port only when it is taken, so nothing changes on a single-user machine. It writes whichever port it got to a per-session file, and clients read it from there:

```
~/.glzr/glazewm/ipc-port-<session id>
```

The session ID is part of the file name because one user can be logged into more than one session at a time, console plus RDP for instance. Without it, the two instances would write over each other and clients would reach the wrong WM. Clients fall back to 6123 when no file is found, so a client talking to an older GlazeWM keeps working.

`wm-platform` gained a small `session_id()` helper for that. On Windows it calls `ProcessIdToSessionId`, on macOS it returns 0 since there is a single GUI session there.

## Compatibility

The port file is new ground for anything outside this repo. Bars, scripts and anything using `glazewm-js` against a hardcoded 6123 keep working on a single-user machine, and reach the first logged-in user's instance otherwise. Zebar needs a matching change, which I will open once this lands.

## Testing

Two Windows users logged in at once, each running their own GlazeWM. Verified that the `Global\` mutex is gone and only the `Local\` one exists, that each session gets its own port file and that it is removed on a clean exit, and that `glazewm-cli query workspaces` reaches the right instance.

`cargo fmt --check` and `cargo clippy --all-targets --all-features -- -D warnings` both pass.

macOS is untested; the change there is limited to the new `session_id()` stub.

## Known trade-off

If the WM is killed rather than exited, the port file is left behind and a client will try a dead port. The user-visible error is the same one they already get when GlazeWM is not running, so this is not a regression, but there is a small new window where Windows could have recycled that ephemeral port for something else. Writing the pid alongside the port and checking it would close the gap; happy to add that if you would rather have it.

## Note on the first commit

This branch sits on top of #1436, a one-line clippy fix, because the lint CI job fails on `main` without it. I will rebase this once that merges.


herlondf:
Pushed one more commit: the WM now exports `GLAZEWM_IPC_PORT` to anything it starts from `startup_commands`.

The port file on its own leaves a bar with a chicken-and-egg problem — it has to work out its own session ID before it can build the file name, and on Windows that means shelling out to `qwinsta` or PowerShell. With the variable it just reads its environment. It came out of the discussion on glzr-io/zebar#296, where the Zebar side of this ended up being the wrong place to solve it.

The variable is set on the WM's own environment rather than per-child, because `ShellExecuteExW` hands the child a copy of ours and takes no environment block of its own. It's a one-shot write during startup, before `startup_commands` run, but it is still a process-global mutation, so shout if you'd rather have it somewhere else.

